PR #1, opened 2023-03-10
moneymoov api client

donalnofrixion:
First draft of the MoneyMoovApiClient.

1. Created a custom hook `usePaymentRequests` for use by a React component. Incrementing the page number will automatically fetch that page.
2. Created `PaymentRequestClient` with support for GET, DELETE and VOID. More to follow.
3. Created `MoneyMoovApiClient` class in typescript for non React clients.
4. Added PaymentRequest types and enums.
5. Created base ApiClient.


sauravmaiti22 (approved):

src/api/ApiClient.ts
```diff
@@ -0,0 +1,79 @@
...
+            headers: {
+            "Content-type": "application/json",
+            "Authorization": `Bearer ${token}`
+            }
+        });
+    
+        let json: string | any;
+
+        try{
+            json = await response.json();
```
Publicker, line 54: What about parsing the response with an expected type (`T`)?

And using `zod` for parsing too.

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+import { PaymentRequestPageResponse } from "./types/ApiResponses";
+import { ApiError } from "./types/ApiResponses";
+
+class MoneyMoovApiClient {
+  apiUrl: string;
+  token: string;
+
+  /**
+   * @param apiUrl The api base url
+   * @param token The token
```
Publicker, line 11: Can we specify a bit more what this token is, and maybe where to get it too?

I believe this will be very beneficial for people using the library, as we all know, there's loads of people who don't exactly read docs, so might be good to have it as IntelliSense here too.

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
+import { deletePaymentRequest, getPaymentRequests, voidPaymentRequest } from "./PaymentRequestClient";
+import { PaymentRequestPageResponse } from "./types/ApiResponses";
+import { ApiError } from "./types/ApiResponses";
+
+class MoneyMoovApiClient {
+  apiUrl: string;
+  token: string;
+
+  /**
+   * @param apiUrl The api base url
```
Publicker, line 10: What about specifying something like "MoneyMoov Api base url" and give an example as it's in the `getPaymentRequests` method.

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+    apiUrl: string,
+    token: string,
+    page: number = 1,
+    size: number = 20
+  ): Promise<{
+    data?: PaymentRequestPageResponse | undefined;
+    error?: ApiError | undefined;
+  }> => {
+    return await getPaymentRequests(apiUrl, token, page, size);
+  };
```
Publicker, line 36: Not sure if we need to pass the `apiUrl` and `token` as a param. Couldn't we use `this.apiUrl` here?

 I think both values should be accessible since it's a method within the same class (`MoneyMoovApiClient`).

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Deletes a Payment Request
+   * @param baseApiUrl The base api url. E.g https://api.nofrixion.com/api/v1
+   * @param token The auth token.
+   * @param paymentRequestId The Payment Request Id
+   * @returns
+   */
+  deletePaymentRequest = async (
+    baseApiUrl: string,
+    token: string,
```
Publicker, line 47: Same comment of use `this`

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Voids a Payment Request
+   * @param baseApiUrl The base api url. E.g https://api.nofrixion.com/api/v1
+   * @param token The auth token.
+   * @param paymentRequestId The Payment Request Id
+   * @returns
+   */
+  voidPaymentRequest = async (
+    baseApiUrl: string,
+    token: string,
```
Publicker, line 64: Same comment of use `this`

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+    paymentRequestId: string
+  ): Promise<{
+    error?: ApiError | undefined;
+  }> => {
+    return await deletePaymentRequest(baseApiUrl, token, paymentRequestId);
+  };
+
+  /**
+   * Voids a Payment Request
```
Publicker, line 56: Might exceed PR's porpoise, but I believe that'd be great if we could add better description for the methods.

For example, in void we'd clarify what "voiding" a payment request means.

Something like

> This function voids a transaction made through a payment request, canceling or invalidating it after it has been initiated but before it has been processed.
> The voided transaction is effectively erased from the record as if it never happened, and any associated funds are returned to their original source.
> This function is useful for correcting mistakes, such as entering the wrong amount or sending a payment to the wrong recipient.
> Note that voiding a transaction is different from issuing a refund, which occurs after the transaction has been processed and involves returning funds to the customer's account. 


I know this might be too much, but I curious to hear your thoughts on this.

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+   * @param baseApiUrl The base api url. E.g https://api.nofrixion.com/api/v1
+   * @param token The auth token.
+   * @param paymentRequestId The Payment Request Id
+   * @returns
```
Publicker, line 60: What about returning a `boolean` just to get a confirmation that the API call went through?

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+  }> => {
+    return await getPaymentRequests(apiUrl, token, page, size);
+  };
+
+  /**
+   * Deletes a Payment Request
+   * @param baseApiUrl The base api url. E.g https://api.nofrixion.com/api/v1
+   * @param token The auth token.
+   * @param paymentRequestId The Payment Request Id
+   * @returns
```
Publicker, line 43: What about returning a boolean just to get a confirmation that the API call went through?

src/api/MoneyMoovApiClient.ts
```diff
@@ -0,0 +1,73 @@
...
+  apiUrl: string;
+  token: string;
+
+  /**
+   * @param apiUrl The api base url
+   * @param token The token
+   */
+  constructor(apiUrl: string, token: string) {
+    this.apiUrl = apiUrl;
+    this.token = token;
```
Publicker, line 15: What do you think of making the token the first argument so `apiUrl` could be optional and default it to the prod api url?

It might not be much but will mean one parameter less for them?

src/api/hooks/usePaymentRequests.ts
```diff
@@ -0,0 +1,36 @@
...
+    
+  const [paymentRequests, setPaymentRequests] = useState<PaymentRequest[]>([]);
+  const [pageNumber, setPageNumber] = useState(1);
+  const [totalPages, setTotalPages] = useState(1);
+
+    useEffect(() => {
+      
+      const fetchPaymentRequests = async () => {
+        
+        const response = await getPaymentRequests(apiUrl, token, page, pageSize);
```
Publicker, line 15: I won't encourage to use use the methods of a specific client, in this case `PaymentRequestClient`, as we expect people to use  the methods of this from the general wrapper, `MoneyMoovApiClient`.

I imagine the usage to be more more like (syntax could be wrong, this is just to demostrate):
```
const moneyMoov = new MoneyMoovClient("eyJhbGc...") // Use token

const response = await moneyMoov.getPaymentRequests(page, pageSize);
```

src/api/types/ApiResponses.ts
```diff
@@ -0,0 +1,151 @@
+export interface PaymentRequestPageResponse extends PageResponse<PaymentRequest> {}
+
```
Publicker, line 2: Loved this! I believe this'd be the place to the `zod` types if we decide to move in that way

donalnofrixion:
Closing in favour of feature/MOOV-1523-moneymoov-api-client-2 PR

PR #2, opened 2023-03-13
moov 1523 moneymoov api client 2

donalnofrixion:
I abandoned the previous PR in favour of this one.

Created a `MoneyMoovApiClient` class for access to all MoneyMoov apis.

`const client = new MoneyMoovApiClient(apiUrl, token);`
`const response = await client.**PaymentRequests**.getAll(page, pageSize);`

Created a PaymentRequestClient class for access to PaymentRequest specific actions.

`const client = new PaymentRequestClient(apiUrl, authToken);`
`const response = await client.getAll(page, pageSize);`

Decided to use the Axios HttpClient (https://axios-http.com/docs/intro)
Advantages are strongly typed responses, automatic form data serialization and simple error handling.


src/api/clients/BaseApiClient.ts
```diff
@@ -0,0 +1,99 @@
...
+import { ApiError } from "../types/ApiResponses";
+import { HttpMethod } from "../types/Enums";
+
+export abstract class BaseApiClient{
+
+    authToken: string;
+
+    constructor(authToken: string){
+        this.authToken = authToken;
+    }
```
Publicker, line 11: What do you think about including the url to this class too? Seems like this base Api Client will always be used with the URL and the inherited classes could re-use the constructor.

sauravmaiti22 (approved):

src/api/clients/BaseApiClient.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            console.log(`Requesting: ${method} ${url}`);
+
+            let contentType = 'application/json';
+
+            // Send form encoding on POST and PUT
+            // Axios will automatically serialize the postData object to form urlencoded format
+            if (method === HttpMethod.POST || method === HttpMethod.PUT){
+                contentType = 'application/x-www-form-urlencoded';
+            }
```
sipsorcery, line 58: Might be easier to use JSON encoding?

donalnofrixion: Currently not all of our controller actions support it. Axios will take the payload and serialize it for a form.

Publicker (approved):
Nothing much to say honestly! I quite like the approach.

The only thing that I'll say, but just to keep in mind is that we're still dealing with a non-real parsing here, meaning that when we receive data from Axios, we're assuming the types rather than validating them since Axios doesn't do this for us.

I'm not saying this has to be in this PR, but at least have it in mind to discuss it at certain point.

PR #6, opened 2023-03-14
Add chip component

Publicker:
This PR adds the `Chip` component to the project, which will be used in the `Row component`. Additionally, the Storybook stories have been modified to include the Chip component.

The stories for the Chip component include:
- Regular: A chip with the label set to "Tag 1"
- Row of tags: Three chips displayed horizontally in the same div

These stories will help us visualise and test the Chip component.

**Screenshot**:

| Regular | Row of tags |
|---------|-------------|
|     Chip component    |       Row of tags      |

### DIFF
--- a/src/components/Chip/Chip.stories.tsx
+++ b/src/components/Chip/Chip.stories.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+
+import Chip from './Chip';
+
+export default {
+  title: 'Chip',
+  component: Chip,
+  argTypes: {
+    label: { control: 'text' },
+  },
+} as ComponentMeta<typeof Chip>;
+
+const Template: ComponentStory<typeof Chip> = (args) => <Chip {...args} />;
+
+export const Regular = Template.bind({});
+Regular.args = {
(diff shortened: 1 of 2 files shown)


donalnofrixion (approved):

src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+  label: { control: { disable: true } },
+};
+
+RowOfTags.decorators = [
+  () => (
+    <div className="space-x-1">
+      <Chip {...(RowOfTags.args, { label: 'Tag 1' })} />
+      <Chip {...(RowOfTags.args, { label: 'Another tag' })} />
+      <Chip {...(RowOfTags.args, { label: 'Third tag' })} />
+    </div>
```
sipsorcery, line 32: Work in prorgress or jsut a demo?

Publicker: The files that are specifically for Storybook's stories (files that end with .stories.tsx) are not just for development purposes, but they are also useful for demonstrating how things will look; so it seemed like a good idea to show how multiple chips will look, since that might be a common use case.

sipsorcery (approved):

PR #8, opened 2023-03-15
[MOOV-1532]: Added payment request status badge component

agranillonf:
Added a badge component to visually represent the three statuses of payment requests:

- Unpaid
- Partially paid
- Fully Paid


README.md
```diff
@@ -4,12 +4,12 @@ NoFrixion Web Components is a collection of web components that can be used to b
 
 ## 💻 Requirements
 
-- [Node.js](https://nodejs.org/en/)
+- [Node.js 16.x](https://nodejs.org/en/blog/release/v16.16.0)
```
Publicker, line 7: Good catch! Thanks!

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+
+import StatusBadge from './StatusBadge';
+
+export default {
+  title: 'StatusBadge',
+  component: StatusBadge,
+  argTypes: {
+    status: { control: 'inline-radio', options: ['UNPAID', 'PARTIAL', 'PAID'] },
```
Publicker, line 10: In the Figma seems like the Status Badge component is always in capital letters. Can we change the component to uppercase the `status` param and pass a normal capitalised word to see if works correctly?

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+  },
+} as ComponentMeta<typeof StatusBadge>;
+
+const Template: ComponentStory<typeof StatusBadge> = (args) => <StatusBadge {...args} />;
+
+export const UnpaidBadge = Template.bind({});
+UnpaidBadge.args = {
+  status: 'UNPAID',
```
Publicker, line 18: Might be better to have all 3 statuses in an object (emulating an enum) so this won't get repetitive and easier if there're new states.

I was thinking in something like 
```jsx
const statuses = {
  unpaid: "unpaid",
  partialPaid: "partial",
  paid: "paid"
};
```

And then you'd use it as:
```jsx
export const PartiallyPaidBadge = Template.bind({});
PartiallyPaidBadge.args = {
  status: statuses.partial,
};
```

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,38 @@
+import unpaidIcon from '../../assets/images/paymentrequest_status_unpaid.svg';
+import paidIcon from '../../assets/images/paymentrequest_status_paid.svg';
+import partiallyPaidIcon from '../../assets/images/paymentrequest_status_partial.svg';
+
+interface StatusBadgeProps {
+  status: 'UNPAID' | 'PARTIAL' | 'PAID';
```
Publicker, line 6: Seems odd to have to send a prop all in capital letters. If we add this to the other comment, the prop value can become lowercased :D

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,38 @@
...
+    default:
+      return unpaidIcon;
+  }
+};
+
+const StatusBadge = ({ status }: StatusBadgeProps) => {
+  return (
+    <div
+      className={`rounded-full py-1 px-2 space-x-2 inline-block ${
+        status === 'PAID' ? 'bg-[#D8F2EA] text-[#008055]' : 'bg-[#F1F3F4] text-[#00264D]'
```
Publicker, line 25: I def forgot to mention this before as it didn't come needed until now, sorry about that.

Anyways, I strongly suggest, from now on to use [cva](https://www.npmjs.com/package/class-variance-authority) (aka `class-variance-authority`) in these cases.

Here's a great video showing what it does and how improves the overall code: https://youtu.be/B6FrDu2Qbt0.

(if you want to go straight to the juice, you can just forward to [11:45](https://youtu.be/B6FrDu2Qbt0?t=705))

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,38 @@
...
+    >
+      <img className="h-3 w-3 float-left mt-0.5" src={getIconForStatus(status)} alt={status} />
+      <span className="text-xs font-normal uppercase float-right">{status}</span>
+    </div>
+  );
+};
+
+StatusBadge.componentProps = {
+  status: String,
+};
```
Publicker, line 36: This is not needed as this won't be a web component

src/web-components/StatusBadge/WebStatusBadge.stories.tsx
Publicker, line 1: Same here, not needed as won't be a web component.

src/web-components/StatusBadge/WebStatusBadge.tsx
Publicker, line 1: And same here

Publicker (changes requested):

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const StatusBadge = ({ status }: StatusBadgeProps) => {
+  return (
+    <div
+      className={`rounded-full py-1 px-2 space-x-2 inline-block ${
+        status === 'PAID' ? 'bg-[#D8F2EA] text-[#008055]' : 'bg-[#F1F3F4] text-[#00264D]'
+      }`}
+    >
+      <img className="h-3 w-3 float-left mt-0.5" src={getIconForStatus(status)} alt={status} />
```
Publicker, line 28: This image doesn't seem to be vertically aligned. The `mt-0.5` could give the sense of aligned, but it's not properly doing it,

src/components/StatusBadge/StatusBadge.tsx
Publicker, line 25: Could you add the `#00264D` color to the as a tailwind color?

It seems that will be used in different places, and even in the Figma document the color was named "Default text" as it's quite used.

pabloi-nf:
Axel,

Looking at the different states it made it clear that it needed better differentiation between the different states.
There is a new color and icon version for the 3 states in Figma. Let me know if there is any question. 
image

src/components/StatusBadge/StatusBadge.tsx
```diff
@@ -1,12 +1,28 @@
...
+    intent: 'unpaid',
+  },
+});
+
+type Status = 'UNPAID' | 'PARTIAL' | 'PAID';
 
 interface StatusBadgeProps {
-  status: 'UNPAID' | 'PARTIAL' | 'PAID';
+  status: String;
 }
```
Publicker: I might not have made myself clear with the previous message regarding using upper or lower case and this could've generated a misunderstanding, apologies.

I was thinking in something more like receiving the props in lower case (as usual) and then make the necessary adjustments in the UI.

Here's kind of how I was thinking it:
```jsx
type Status = 'unpaid' | 'partial' | 'paid'; // 1. Options are in lowered case as usual

interface StatusBadgeProps {
  status: Status; // 2. Type is being used (instead of `string`) 
}

const getIconForStatus = (status: Status) => {
  switch (status) {
    case 'partial':
      return partiallyPaidIcon;
    case 'paid':
      return paidIcon;
    case 'unpaid':
    default:
      return unpaidIcon;
  }
};

// 3. Note that the `parseStatus` function won't do anything now because we already got the type from the prop itself (see point 2)

const StatusBadge = ({ status }: StatusBadgeProps) => {
  // 4. No conditional rendering is needed
  return (
    <div className={badge({ intent: status })}> 
      <img className="h-3 w-3" src={getIconForStatus(status)} alt={status} />
      <span className="text-xs font-normal uppercase">{status.toUpperCase()}</span> {/* 5. We are uppercasing the string ONLY for the UI */}
    </div>
  );
```

---

With this changes you then will have to update the story for this component,

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -2,28 +2,35 @@ import React from 'react';
...
+  partial: 'Partial',
+  paid: 'Paid',
+};
 
 export default {
   title: 'StatusBadge',
   component: StatusBadge,
   argTypes: {
-    status: { control: 'inline-radio', options: ['UNPAID', 'PARTIAL', 'PAID'] },
+    status: { control: 'inline-radio', options: [statuses.unpaid, statuses.partial, statuses.paid] },
```
Publicker: With the changes suggested in the comment of the `StatusBadge.tsx`, you could tweak this story to make it a bit better.

Here's my suggestion:
```jsx
// 1. Statues now will be lowered-case
const statuses = {
  unpaid: 'unpaid',
  partial: 'partial',
  paid: 'paid',
};

export default {
  title: 'StatusBadge',
  component: StatusBadge,
  argTypes: {
    // 2. Object.values() will save you from hand-writing the options 🤘🏻
    status: { control: 'inline-radio', options: Object.values(statuses) },
  },
} as ComponentMeta<typeof StatusBadge>;
```

src/components/StatusBadge/StatusBadge.stories.tsx
```diff
@@ -2,28 +2,35 @@ import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 
 import StatusBadge from './StatusBadge';
+import { StatusBadgeProps } from '@storybook/addon-interactions/dist/ts3.9/components/StatusBadge/StatusBadge';
```
Publicker, line 5: This import could be removed with the suggested changes.

Publicker (changes requested):

Publicker (approved):
Lovely stuff!

PR #9, opened 2023-03-15
Paging component

donalnofrixion:
Added new paging component called Pager.tsx


arifmatin (approved):

src/components/Pager/Pager.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+  component: Pager,
+  argTypes: {
+    pageSize: { control: 'number' },
+    totalRecords: { control: 'number' },
+  },
+} as ComponentMeta<typeof Pager>;
+
+const Template: ComponentStory<typeof Pager> = (args) => <Pager {...args} />;
+
+export const Primary = Template.bind({});
```
Publicker, line 17: What do you think of calling this `Showcase` or `Playground`, the Primary nomenclature was used as an example. 

A pattern that we could follow is to always create a `Playground` or `Showcase` story (your Primary), and then do a few examples, let's say, with different `pageSize` and `totalRecords`.

Stories that could be good for instance:
- Showcase
- 150 items
- 2418 items
- 30 items per page
- And so on...

Kind of thinking the stories as "tests" which will be ultimately "shortcuts" for different cases (different prop values)

Publicker (changes requested):

Publicker:
All the changes worked fine!

However, I found some strange behaviours when changing the values, specially with `pageSize`.

1. `pageSize` doesn't seem to trigger a re-render automatically (seems to change once you change `totalRecords`
2. I faced an issue (check video attached) where I could go to pages after I should've reached the end (probably related with the point 1).

Video showing both behaviours:

https://user-images.githubusercontent.com/52673485/225373828-b6ada448-3149-4a18-9c6b-30f0490a3e7c.mp4

pabloi-nf:
Hi Donal,

Looks great so far! A few comments on style and behaviour:

- The component should be right aligned, so when it changes length because of longer numbers, the arrows don't change position.
- There must be 16px distance between the numbers and the arrows.
- There must be a hover style for the arrows. The color should change to #454D54. I added this on the Figma file too.

Thanks!

src/components/Pager/Pager.tsx
```diff
@@ -59,31 +62,43 @@ const Pager = ({ pageSize, totalRecords, onPageChange }: PagerProps) => {
...
+          width="7"
+          height="13"
+          viewBox="0 0 7 13"
+          fill="none"
+          xmlns="http://www.w3.org/2000/svg"
           onClick={() => decrementPageNumber()}
-        />
+        >
+          <path d="M6.5 0.5L0.5 6.5L6.5 12.5" stroke-linecap="round" stroke-linejoin="round" />
+        </svg>
```
Publicker: While this approach of directly embedding the SVG content works, it's not the most ideal practice.

In general, it's preferable to keep assets as separate files, and only use the svg code directly for uncommon modifications (e.g., altering path values for animation purposes).

For this case, I recommend continuing to import the SVG from the files and applying the fill class after the import. You can find more information on using the fill class in the [Tailwind CSS documentation](https://tailwindcss.com/docs/fill).

By following this approach, we can maintain a cleaner and more organised codebase.

---

Please feel free to ping me in Slack if you've any question!

src/components/Pager/Pager.tsx
```diff
@@ -59,31 +62,43 @@ const Pager = ({ pageSize, totalRecords, onPageChange }: PagerProps) => {
...
+          width="8"
+          height="13"
+          viewBox="0 0 8 13"
+          fill="none"
+          xmlns="http://www.w3.org/2000/svg"
           onClick={() => incrementPageNumber()}
-        />
+        >
+          <path d="M1 12.5L7 6.5L0.999999 0.500001" stroke-linecap="round" stroke-linejoin="round" />
+        </svg>
```
Publicker: Same comment for this SVG

Publicker (changes requested):

pabloi-nf (approved):

src/components/Pager/Pager.tsx
donalnofrixion: @Publicker I can't seem to get the 'hover:stroke' working when using an 'img' as opposed to an 'svg'. When you're free can you give me a shout to discuss?

Publicker (approved):
Great stuff!

PR #10, opened 2023-03-15
Add Row component 

Publicker:
This pull request introduces the `Row` component, intended to be used by the Table component in upcoming implementations. A Storybook story has been added for testing and visualisation purposes.

Changes include:

1. Creation of the Row component, designed to function as a <tr> HTML element within a table structure.
2. Inclusion of a temporary table as a Storybook decorator for enhanced display, to be replaced with the proper Table component in the future.

It should be noted that the `Status` and `Ellipsis` components have not been included in this PR, as they are currently under development separately.

--- 

Added the following stories with corresponding links:

- [Row / Example Playground](https://6410c09a16b1ae3112580b1c-fycgrmzdwz.chromatic.com/?path=/story/row--playground)
- [Row / Example Today](https://6410c09a16b1ae3112580b1c-fycgrmzdwz.chromatic.com/?path=/story/row--example-today)
- [Row / Example Yesterday](https://6410c09a16b1ae3112580b1c-fycgrmzdwz.chromatic.com/?path=/story/row--example-yesterday)
- [Row / Example Two Days Ago](https://6410c09a16b1ae3112580b1c-fycgrmzdwz.chromatic.com/?path=/story/row--example-two-days-ago)
- [Row / Example Two Years Ago](https://6410c09a16b1ae3112580b1c-fycgrmzdwz.chromatic.com/?path=/story/row--example-two-years-ago)
- [Row / Example GBP](https://6410c09a16b1ae3112580b1c-fycgrmzdwz.chromatic.com/?path=/story/row--example-gbp)

### DIFF
--- a/src/components/Row/Row.tsx
+++ b/src/components/Row/Row.tsx
@@ -1,0 +1,43 @@
+import { formatAmount, formatDate } from '../../utils/formatters';
+import Chip from '../Chip/Chip';
+import Contact from '../Contact/Contact';
+
+interface RowProps {
+  status: 'paid' | 'partial' | 'unpaid';
+  createdAt: Date;
+  contact: LocalContact;
+  amount: number;
+  currency: 'EUR' | 'GBP';
+  tags: string[];
+}
+
+const Row = ({ status, createdAt, contact, amount, currency, tags }: RowProps) => {
+  return (
+    <tr className="border-b border-[#F1F2F3]">
+      {/* TODO: Replace status text for <Status> component */}
+      <td className="py-3">{status.charAt(0).toUpperCase() + status.slice(1)}</td>
+
+      <td className="text-sm">{formatDate(createdAt)}</td>
+
+      <td>
+        <Contact {...contact} />
+      </td>
+
+      <td className="text-right truncate">
+        <span className="font-medium">{formatAmount(amount)}</span>
+      </td>
+
+      <td>
+        <span className="ml-5 text-greyText">{currency}</span>
+      </td>
+
+      <td className="space-x-1 text-right">
+        {tags.map((tag) => (
+          <Chip label={tag} />
+        ))}
+      </td>
+    </tr>
+  );
+};
+
+export default Row;
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,9 @@
   "prettier.configPath": ".prettierrc",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
+  "files.associations": {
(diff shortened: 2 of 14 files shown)


donalnofrixion (approved):

arifmatin (approved):

src/components/Row/Row.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface RowProps {
+  status: 'paid' | 'partial' | 'unpaid';
+  createdAt: Date;
+  contact: LocalContact;
+  amount: number;
+  currency: 'EUR' | 'GBP';
+  tags: string[];
+}
+
+const Row = ({ status, createdAt, contact, amount, currency, tags }: RowProps) => {
```
donalnofrixion, line 14: Just thinking. Would it not make more sense to call this PaymentRequestRow and pass a PaymentRequest into it? In future the PaymentRequest Details will be triggered here? I don't think it makes sense to create a resuable row component at this level. Other views will probably have completely different fields/details.

donalnofrixion: Fair enough. We should probably look at trimming down our ApiRsesponses as well. I created them as dupes of the api versions for starters.

Publicker: Yep. That's why I started creating "Local" types (i.e `LocalCurrency`), following the advice from our [bible](https://martinfowler.com/articles/modularizing-react-apps.html).

In a near future we'd find ourselves mapping the API types to the Local ones.

---

Btw, the renaming is already up🤘🏻

pabloi-nf (changes requested):
image

Comparing the Figma file with the HTML I see some small things to fix:

- Font size for date and customer name should be 13px.
- Distance between customer name and customer email should be smaller.
- Currency text should be 14px.
- Apply Inter font variants for characters 1, 4,  and 9.

Publicker:
> image
> 
> Comparing the Figma file with the HTML I see some small things to fix:
> 
> * Font size for date and customer name should be 13px.
> * Distance between customer name and customer email should be smaller.
> * Currency text should be 14px.
> * Apply Inter font variants for characters 1, 4,  and 9.

@pabloi-nf All these items were attended, please take a look at the new version [here](https://6410c09a16b1ae3112580b1c-hzqjaqchlu.chromatic.com/?path=/story/payment-request-row--playground).

Publicker:
@pabloi-nf Hover state & onClick prop added too!

Check it in [here](https://6410c09a16b1ae3112580b1c-rdmzvixuof.chromatic.com/?path=/story/payment-request-row--playground)!

pabloi-nf (approved):

PR #12, opened 2023-03-16
[MOOV-1531]: Added Payment Request Row Action Menu

agranillonf:
Created payment request row action menu component, with the following optional actions:

- onDuplicate: Payment request duplication event handler
- onCopyLink: Payment request link copy event handler
- onDelete: Payment request deletion event handler

Also added both the StatusBadge and the ActionMenu to the PaymentRequestRow component.


src/components/PaymentRequestActionMenu/PaymentRequestActionMenu.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+  title: 'Payment Request Row Action Menu',
+  component: PaymentRequestActionMenu,
+} as ComponentMeta<typeof PaymentRequestActionMenu>;
+
+const Template: ComponentStory<typeof PaymentRequestActionMenu> = (args) => <PaymentRequestActionMenu {...args} />;
+
+export const ActionMenuTemplate = Template.bind({});
+ActionMenuTemplate.args = {
+  onDuplicate: () => {
+    alert('You clicked Duplicate!');
```
donalnofrixion, line 16: You could replace this with action('You clicked Duplicate!') and it will show in the Actions pane

donalnofrixion (approved):

src/components/PaymentRequestActionMenu/PaymentRequestActionMenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+      intent: {
+        neutral: [],
+        negative: ['text-red-700'],
+      },
+    },
+    defaultVariants: {
+      intent: 'neutral',
+    },
+  },
+);
```
Publicker, line 21: We could remove this as no intents are being used.

Maybe keep the default classes in a const variable, and reuse it.

Something like:
```jsx
// ..
const itemClassnames = "group text-xs leading-none rounded-1 flex items-center h-6 py-4 pr-4 relative select-none outline-none cursor-pointer data-[highlighted]:text-greyText"
// ...
```

And then use it as:
```jsx
{/* ... */}

<DropdownMenu.Item className={itemClassnames} onClick={onDuplicate}>

{/* ... */}

<DropdownMenu.Item className={itemClassnames} onClick={onCopyLink}>

{/* ... */}

<DropdownMenu.Item className={itemClassnames} onClick={onDelete}>

{/* ... */}

```

src/components/PaymentRequestActionMenu/PaymentRequestActionMenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+const PaymentRequestActionMenuItemContent = ({ label, iconSource }: PaymentRequestActionMenuItemContentProps) => {
+  return (
+    <>
+      <div className="pl-4 pr-2">
+        <img src={iconSource} alt={label} />
+      </div>
+      {label}
+    </>
```
Publicker, line 41: Maybe Pablo I. would leave this message, but I feel like the items needs a kind of hover to make them feel better.

Do you know if there's any design about this?

src/components/PaymentRequestRow/PaymentRequestRow.tsx
```diff
@@ -39,6 +56,8 @@ const Row = ({ status, createdAt, contact, amount, currency, tags, onClick }: Pa
         {tags.map((tag) => (
           <Chip label={tag} />
         ))}
+
+        <PaymentRequestActionMenu onDuplicate={onDuplicate} onCopyLink={onCopyLink} onDelete={onDelete} />
```
Publicker, line 60: In my opinion this should go in a different `<td>` (you will need to add a new `<th>` in the Row to achieve this too)

Publicker (changes requested):
### Propagation issue

There's an issue with this component and the events. 

If you go to Storybook and try this component inside of the `Row` you will see that each time you click on a `ActionMenuItem` it will not only trigger the `Item` event, but also the `onClick` from the `Row`. This is due to a  propagation issue.

Here's an interesting article explaining this: https://www.robinwieruch.de/react-event-bubbling-capturing/

Here is the demo on the issue:

https://user-images.githubusercontent.com/52673485/225904793-eef61eb7-45c6-49b5-ae73-383f747cb6a6.mp4

----

### Solution

To solve this issue, I think we could create a function in the `PaymentRequestActionMenu` component like:

```jsx
const handleClick = (e: React.MouseEvent<HTMLDivElement>, functionToRun: () => void) => {
  functionToRun();

  e.stopPropagation();
};
```

And then use it as:

```jsx

{/* ... */}

<DropdownMenu.Item
  className={itemClassnames}
  onClick={(e) => {
    handleClick(e, onDuplicate);
  }}
>
  <PaymentRequestActionMenuItemContent label="Duplicate" iconSource={copyIcon} />
</DropdownMenu.Item>

{/* ... */}

<DropdownMenu.Item
  className={itemClassnames}
  onClick={(e) => {
    handleClick(e, onCopyLink);
  }}
>
  <PaymentRequestActionMenuItemContent label="Copy payment link" iconSource={linkIcon} />
</DropdownMenu.Item>

{/* ... */}

<DropdownMenu.Item
  className={itemClassnames}
  onClick={(e) => {
    handleClick(e, onDelete);
  }}
>
  <PaymentRequestActionMenuItemContent label="Delete" iconSource={trashIcon} />
</DropdownMenu.Item>

{/* ... */}
```

pabloi-nf (changes requested):
Axel,

There are some sizes to fix:
- All icons should be 16x16px.
- Padding all around should be 16px. Right now it looks like left and right margins are too small compared with top and bottom.

Besides that, I made a change on the 3 dots icon color. The default state should be #8F99A3, and the current color should be only on hover.
Let's also change the "delete" text color to #F32448, to make it more clear.

Thanks

src/components/PaymentRequestActionMenu/PaymentRequestActionMenu.tsx
agranillonf: It does have a hover effect. The class is assigned to the parent element.

donalnofrixion (approved):

pabloi-nf (changes requested):
Looking good!
A few things to adjust:
- Let's put a hover color to the delete text. t should be #FF6681.
- Did the animation changed? Now it looks like it's growing up. I think that is too much. The previous version only had a moving up animation as far as I remember, which was much more subtle.
- The margins around are still different than the design. There is too much white space at the sides and too little top and bottom. It seems there is a 5px margin all around, and then 16px at the left side of the icon, but no margin on top of it.

It should look like this
image

Not like this
image

pabloi-nf (approved):

arifmatin (approved):

PR #13, opened 2023-03-17
Feature/moov 1563 column sort

donalnofrixion:


src/components/ColumnHeader/ColumnHeader.tsx
```diff
@@ -0,0 +1,81 @@
...
+import classNames from 'classnames';
+
+export enum SortDirection {
+  NONE = 'NONE',
+  ASC = 'ASC',
+  DESC = 'DESC',
+}
+
+export interface SortEvent {
+  name: string;
```
Publicker, line 11: I don't think there's a need of sending the `name` on this event. I believe this will be an unused prop. I'd rather just straight send the direction.

The reason behind this is that when the component is being used, you'd know which column you're controlling, because you're specifying a particular function.

Here's an example of usage of this component without sending the `name`:

```jsx
const Table = ({prop1, prop2}: TableProps) => {
  const [statusSortDirection, setStatusSortDirection] = useState<SortDirection>(SortDirection.NONE);
  const [createdSortDirection, setCreatedSortDirection] = useState<SortDirection>(SortDirection.NONE);

  return (
    <>
      {/* ... */}

      <ColumnHeader name="Status" onSort={setStatusSortDirection} />

      <ColumnHeader name="Created" onSort={setCreatedSortDirection} />

      {/* ... */}
    <>
  )
}

```
But sending the name in the sort event will only add boilerplate code, here's the same example but sending the name in the event:

```jsx
const Table = ({prop1, prop2}: TableProps) => {
  const [statusSortDirection, setStatusSortDirection] = useState<SortDirection>(SortDirection.NONE);
  const [createdSortDirection, setCreatedSortDirection] = useState<SortDirection>(SortDirection.NONE);

  return (
    <>
      {/* ... */}

      <ColumnHeader name="Status" onSort={({name, sortDirection}) => setStatusSortDirection(sortDirection))} />

      <ColumnHeader name="Created" onSort={({name, sortDirection}) => setCreatedSortDirection(sortDirection))} />

      {/* ... */}
    <>
  )
}
```

src/components/ColumnHeader/ColumnHeader.tsx
```diff
@@ -0,0 +1,81 @@
...
+      setSortDirection(SortDirection.ASC);
+    } else if (sortDirection === SortDirection.ASC) {
+      setSortDirection(SortDirection.DESC);
+    } else {
+      setSortDirection(SortDirection.NONE);
+    }
+  };
+
+  return (
+    <>
```
Publicker, line 41: Fragment is not needed as this is returning just one component

src/components/ColumnHeader/ColumnHeader.tsx
```diff
@@ -0,0 +1,81 @@
...
+            width="10"
+            height="6"
+            viewBox="0 0 10 6"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path d="M9 1L5 5L1 1" strokeLinecap="round" strokeLinejoin="round" />
+          </svg>
+        </div>
+      </div>
```
Publicker, line 76: I've experienced a strange feeling when testing the component so I started to investigate and found a few things with it.

Not 100% relevant to the code comment, but I think it'll be worth to mention so in the future we'd try to test different things - here're few things I noticed during the test:
1. Cursor pointer is lost when moving in between the label and the arrows
1. Arrows have hover effect, but no different action than clicking in the label. The hover effect made me think I could set a particular order instead of alternating, but this it wasn't the case.
1. Arrows are too close
1. If I started to alternate the status relatively quick, the label will be selected (as any other text would), but it felt kind of wrong, because it's more a button rather than text.

Anyway, with these test result, my first step was to turn on the outlines in Storybook. Here's how it looked:
Storybook ColumnHeader component outline lines

As you can see the outlines shows that there's an issue with the component size. 

After looking at the code, I think we'd do some changes to improve the items mentioned above. Here's some code which possibly will possible fix these issues (with comments explaining why)

```jsx
<div
  // 1. Changed classnames to fix cursor pointer issue & fix sizes shown in the Storybook outline
  // 1.1. Removed height & width as there's no need to set them. This should be automatically set.
  // 1.2. Changed grid system to use flexbox instead. As mentioned in the other comment, flexbox should be used when the layout is for one dimension.
  // 1.3. Added transition to hover state to make it smoother.
  // 1.4. Added items-center to align the text & arrows vertically.
  // 1.5. Removed mt-1 & ml-1 as this should be handled by the parent component.
  className="inline-flex text-sm text-greyText cursor-pointer transition hover:text-controlGreyHover items-center"
  onClick={doSort}
>
  {/*
    4. Changed classnames to make text unselectable
    4.1. Changed `div` to `span` as this is a text element.
    4.2. Added `select-none` to make the text unselectable.
    4.3. Removed .toUpperCase() and added `uppercase` to make the text uppercase. (less computationally expensive)
  */}
  <span className="select-none uppercase">{label}</span>

  {/*
    3. Changed classnames to improve distance between arrows
    3.1. Removed `mt-1` as this wasn't properly aligning vertically. And this is now done (see 1.4. above)
    3.2. Added `space-y-1` to add some vertical distance between the arrows.
  */}
  <div className="ml-2.5 space-y-1">
    {/*
      2. Changed classnames to remove hover state as this was causing the impression that the arrows were doing a different action
      2.1. Removed `hover:stroke-controlGreyHover`.
    */}
    <svg
      className={classNames({
        'stroke-controlGreyHover': sortDirection === SortDirection.ASC,
        'stroke-controlGrey': sortDirection === SortDirection.NONE || sortDirection === SortDirection.DESC,
      })}
      width="10"
      height="6"
      viewBox="0 0 10 6"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
    >
      <path d="M1 5L5 1L9 5" strokeLinecap="round" strokeLinejoin="round" />
    </svg>
    {/*
      2. Changed classnames to remove hover state as this was causing the impression that the arrows were doing a different action
      2.1. Removed `hover:stroke-controlGreyHover`.
    */}
    <svg
      className={classNames({
        'stroke-controlGreyHover': sortDirection === SortDirection.DESC,
        'stroke-controlGrey': sortDirection === SortDirection.NONE || sortDirection === SortDirection.ASC,
      })}
      width="10"
      height="6"
      viewBox="0 0 10 6"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
    >
      <path d="M9 1L5 5L1 1" strokeLinecap="round" strokeLinejoin="round" />
    </svg>
  </div>
</div>
```

Publicker (changes requested):

src/components/ColumnHeader/ColumnHeader.tsx
donalnofrixion: Thanks @Publicker really appreciate You taking the time to look at this.

Publicker (approved):
Great job! 🚀

pabloi-nf (commented):
Donal, 
I'm looking at the result and I feel there is something odd about the behavior. If I click on the down arrow I expect the column to be sorted in that direction, not to start from the top. I don't think is the wrong behavior, but I feel the UI needs a change to better communicate this.
If you give me some minutes I'm coming with a change proposal.
Thanks

pabloi-nf:
We can keep the behavior but change how we show the arrows.
1. By default the arrows are hidden.
2. When hovering, the arrows are shown and the text changes color.
3. If clicked, the down sorting gets active and the text goes back to the gray color.
4. If clicked again, up sorting gets active.
5. Clicking a third time deactivates the sorting, bringing the title to its initial state.

Color for inactive arrows: #B9BFC6
Color for active arrows: #454D54


image
 
Thanks!

pabloi-nf (approved):

PR #18, opened 2023-03-23
Feature/moov 1527 date range filter

donalnofrixion:


src/components/ui/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { motion } from 'framer-motion';
+import { cva } from 'class-variance-authority';
+import { DateRangePicker as DateRange } from 'rsuite';
+import '../../../rsuite.css';
+import { format } from 'date-fns';
+import isAfter from 'date-fns/isAfter';
+
+const { afterToday } = DateRange;
+
+const pillClasses = 'bg-[#EDF2F7] text-sm whitespace-nowrap border-[1px] border-[#D5DBDD] cursor-pointer';
```
agranillonf, line 15: You could use our custom bg-greyBg class instead of bg-[#EDF2F7]

src/components/ui/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }, [selectRangeText]);
+
+  return (
+    <div className="flex defaultText">
+      <DropdownMenu.Root>
+        <DropdownMenu.Trigger asChild>
+          <div
+            className={classNames(
+              pillClasses,
+              'text-[#00264D] rounded-l-full border-r-0 flex flex-col-2 space-x-2 pl-4 pr-3',
```
agranillonf, line 93: text-[#00264D] -> text-defaultText

src/components/ui/PaymentRequestActionMenu/PaymentRequestActionMenu.tsx
```diff
@@ -7,7 +7,7 @@ import { cva } from 'class-variance-authority';
 import { motion } from 'framer-motion';
 
 const actionItemClassNames =
-  'group text-xs leading-none rounded-1 flex items-center relative select-none outline-none cursor-pointer';
+  'group text-xs leading-none rounded-1 flex items-center relative select-none outline-none cursor-pointer p-y-1';
```
agranillonf, line 10: Did you mean py-1?

agranillonf (approved):

src/index.css
```diff
@@ -85,3 +85,91 @@
...
+  color: var(--rmdp-primary-green);
+}
+
+.green .rmdp-button:not(.rmdp-action-button) {
+  background-color: var(--rmdp-primary-green);
+}
+
+.green .rmdp-button:not(.rmdp-action-button):hover {
+  background-color: var(--rmdp-deselect-green);
+}
```
Publicker, line 175: Can we move this into a different `.css` file which indicates that the styles are dedicated for the calendar?

Something like `DateRangeInput.css` and move it inside of the `DateRangePicker` folder.

src/utils/constants.ts
```diff
@@ -4,3 +4,11 @@ export const apiUrls = {
...
   production: 'https://api.nofrixion.com/api/v1',
 };
+
+export const dateRanges = {
+  today: 'Today',
+  yesterday: 'Yesterday',
+  last7Days: 'Last 7 Days',
+  last30Days: 'Last 30 Days',
+  last90Days: 'Last 90 Days',
+};
```
Publicker, line 14: This logic could be moved all together to use `date-fns` helper functions and this will make easier the behaviour of showing "Today" when someone picked today in the calendar (as the current behaviour is showing "Custom").

src/utils/formatters.ts
```diff
@@ -78,4 +78,28 @@ const formatPaymentRequestSortExpression = (
...
+  date.setDate(date.getDate() - daysToGoBack);
+
+  if (startOfDay) {
+    date.setHours(0, 0, 0, 0);
+  } else {
+    date.setHours(23, 59, 59, 999);
+  }
+
+  return date;
+};
```
Publicker, line 93: This function could be replaced for another `date-fns` helper function called `add`.

The way you'd use the function is:
```jsx
add(yourDateVariable, { days: -4 })
```

Where the `-4` value should be replaced by the amount of days.

Publicker (changes requested):

pabloi-nf (changes requested):
- Is it possible to remove the gray background? Even if this is the color is supposed to be, I feel this component needs to be transparent to allow it to be in more than one place or if the background color changes in the future.

Preset list:
- The font size on the preset options list should be 14px.
- In the same list, the separation between items should be bigger. In the design the line-height of each item is 24px, and they gave 16px in between them. In the HTML the line-height seems to be 14px. You can change this or change the margins, as long as it creates the same visual separation.
- The color of the active text in that list should be #009999.

On the calendar side:
- The date field has a line-height property (1.25rem) that is moving all the text a little up, making it look uncentered. I think if we remove it it should be fixed.
- The color and shadow of the box should be the same as the preset list. We should remove the little arrow on top.
- The icon looks weird on hover. It seems some lines are lighter than others 
image
- The color and style of the arrows to change month should match the Figma file.
- The color of the text is black, but it should be the same dark blue #00264D of the rest of the modules.
- The day names should be gray #73808C, and if possible only 2 letters long. Font size should be 10px.
- The selection row should not have a shadow (or glow?) around. 
image
- The only white day numbers should be the start and end. All others should be #004D40, with a background #D9F2F2.
- Lets use a different bg color for today's number: #EDF2F7.

If there is anything not clear or that can't be done please let me know to find a solution.

Thanks!

pabloi-nf (changes requested):
I can still see a difference between the Figma (left) and the HTML (right).
image

From what I can see, there is an extra 8px margin on top that creates a 24px separation between items, instead of 16px. That makes them look too far apart.
image

There is also too much space on top and bottom. The 16px box padding plus the 8px padding of the items creates too much white space. Let's reduce the top and bottom box padding to 4px. Lets increase left and right margin to 24px, like in the Figma, and remove the "min-width" property, so the box adapts to the longest list element.

Besides this, I noticed a bug with the font size in the future days:
image


Thanks! This should be all

donalnofrixion:
https://user-images.githubusercontent.com/89344328/228559491-ee2e3ccb-421f-4f94-bf53-695a8322b808.mov

arifmatin (approved):

pabloi-nf:
@donalnofrixion, it looks great with the animation. The only thing I'm noticing is that it takes some time to start, creating a lag feeling. I think is because it has an ease-in ease-out. Maybe if we only leave an ease-out effect it would look better.

donalnofrixion:
Hi @pabloi-nf I fixed the animation. Let me know what you think. Thanks

pabloi-nf (changes requested):
Hi Donal, 

I still think it looks off. If you compare the Figma animation (first video) with the HTML, you can see the first one looks smoother.
I think that is because in the Figma we also animate the arrow, the separator line and icon in between fields, while on the HTML all that appears instantly, and it's only the outer border 
that animates, making it look "late" like its catching up the rest.

I'm not sure how you are animating it and if this is possible. If not, I think eliminating the animation it's a better option.

https://user-images.githubusercontent.com/127946572/228776773-c8e4549c-4208-4106-8405-ac642e6e3c27.mov

https://user-images.githubusercontent.com/127946572/228777293-bee90782-5f45-422d-930e-0262c74da4e9.mov

pabloi-nf (approved):
Perfect!

Publicker (approved):

PR #19, opened 2023-03-24
[MOOV-1571]: Integrated payment request Copy link and deletion menus

agranillonf:
Integrated the "Copy link" and "Delete" actions from the payment request row menu with the corresponding API calls.

The following packages were installed to achieve this:
- [HeadlessUI](https://headlessui.com/react/dialog) (@headlessui/react): For a deletion confirmation dialog.
- [React Toastify](https://fkhadra.github.io/react-toastify/introduction) (react-toastify): For general success/error messages.

For @pabloi-nf:
It's worth checking out the text, colors and animations for both the deletion dialog and copy-link toasts. 
Also, for toasts, I implemented the typical four levels of messaging/logging, each one has a background and a text color, which go as follows:
- Informational (info): white background, default text color (#00264D)
- Success (success): white background, default text color (#00264D)
- Warning (warning): #F2EED8 for background, #663300 for text color. Same colors as the "Unpaid" status badge.
- Error (error): #FECACA for background, #7F1D1D for text color. I guess this is the one where I need your input the most.


src/components/functional/PaymentRequestTable/PaymentRequestTable.tsx
```diff
@@ -30,21 +37,72 @@ const PaymentRequestTable = ({ token, apiUrl = 'https://api.nofrixion.com/api/v1
...
+      let deleteError = await deletePaymentRequest(paymentRequestIdToDelete);
+
+      if (deleteError) {
+        makeToast('error', deleteError.title);
+        return;
+      }
+
+      makeToast('success', 'Payment request successfully deleted.');
+
+      await fetchPaymentRequests();
```
donalnofrixion, line 73: This feels a little brute force. Would it be possible to not do this and instead delete the PaymentRequestRow?

agranillonf: Thing is, if we do that, the current page shown will only be able to show 9 items until the user changes pages or refreshes. Do we want that?

package.json
```diff
@@ -28,7 +29,8 @@
     "framer-motion": "^10.8.4",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-to-html-element": "^1.4.7"
+    "react-to-html-element": "^1.4.7",
+    "react-toastify": "^9.1.2"
```
Publicker, line 33: I have the same feeling about if we need toast as of now. Another thing that could be discussed with the design team @pabloi-nf.

src/components/ui/PaymentRequestTable/PaymentRequestTable.tsx
```diff
@@ -9,6 +9,9 @@ interface PaymentRequestTableProps {
   pageSize: number;
   totalRecords: number;
   onPaymentRequestClicked?: (paymentRequest: LocalPaymentRequest) => void;
+  onPaymentRequestDuplicateClicked?: (paymentRequest: LocalPaymentRequest) => void;
+  onPaymentRequestDeleteClicked?: (paymentRequest: LocalPaymentRequest) => void;
+  onPaymentRequestCopyLinkClicked?: (paymentRequest: LocalPaymentRequest) => void;
```
Publicker, line 14: Maybe we can make all of these required? I don't see a reason why they should be optional now.

src/components/functional/PaymentRequestTable/PaymentRequestTable.tsx
```diff
@@ -30,21 +37,72 @@ const PaymentRequestTable = ({ token, apiUrl = 'https://api.nofrixion.com/api/v1
...
+      setPaymentRequestIdToDelete('');
+    }
+
+    setIsDeleteDialogOpen(false);
+  };
```
Publicker, line 78: I'd try to move as much logic from here to the `UI/PaymentRequestTable`. I just wonder if could be possible to manage to get the dialog in the UI component and then execute the request in here once we got the confirmation from wherever it is, Dialog, double click, etc...

Just to make the UI the most testable without the need of connecting to the API.

Publicker (changes requested):

Publicker (approved):

pabloi-nf (changes requested):
Axel, please see the following prototype for toast behaviour and the delete process.
https://www.figma.com/proto/Sc96Upmn3MpOmvM9jgvaUG/Hi-Fi?page-id=506%3A3283&node-id=1506-8934&viewport=224%2C152%2C0.41&scaling=min-zoom&starting-point-node-id=1506%3A8934

I'm taking Pablo's suggestion to keep the delete process inline for row elements. 

About the toast styles for the different type of messages, I'm still working on it to make it consistent. I'll keep you posted.

src/components/ui/PaymentRequestRow/PaymentRequestRow.tsx
```diff
@@ -27,39 +30,101 @@ const Row = ({
...
+
+  const onDeletePaymentRequestClicked = async () => {
+    setIsDeleting(true);
+    animate(`tr.custom-backdrop-blur-${id}`, { opacity: 0.2 }, { duration: 0.2 });
+  };
+
+  const onCancelDeletingPaymentRequestClicked = async () => {
+    setIsDeleting(false);
+    animate(`tr.custom-backdrop-blur-${id}`, { opacity: 1 }, { duration: 0.2 });
+  };
```
Publicker, line 55: Not sure how difficult would this be, but could it be possible to automatically trigger a cancel deleting payment request when clicking outside the row the user wanted to delete?

Don't want to make this ultra complex it's just that I found myself kind of having this thought.

Publicker (changes requested):

src/components/ui/PaymentRequestTable/PaymentRequestTable.tsx
```diff
@@ -23,68 +27,86 @@ const PaymentRequestTable = ({
...
+      <table className="table-fixed text-left w-full">
+        <colgroup>
+          <col />
+          <col />
+          <col />
+          <col />
+          <col />
+          <col />
+          <col className="w-8" />
+        </colgroup>
```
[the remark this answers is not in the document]
agranillonf: It is, because we need to set the width of the last column (the one with the action menu) and, as we're using a table header with colspan 2, there's no other way to set it.

pabloi-nf (changes requested):
Delete process:
- The red and white buttons need to have a hover state. The colors are in the Figma file.
- When the confirm delete buttons are being displayed, all other interactions with the row should be disabled. Right now I can still open the more options menu.
image

Toast:
- The font size should be 14px.
- Line-height should be 18px.
- Margin top and bottom should be 12px.
- Margin left and right should be 24px.
- Lets add a hover effect on the X icon with a #8F99A3 color.
- Box-shadow should be 0px 0px 8px rgba(4, 41, 49, 0.1)
- Box radius should be 8px (I can't confirm how much it is now)

Thanks!

pabloi-nf (approved):

Publicker (approved):

donalnofrixion (approved):

PR #20, opened 2023-03-31
Feature/moov 1597 payment request dashboard

donalnofrixion:


sipsorcery:
If there's 0 matches on a filter type the tabbing doesn't work. In the below example the last tab selected was "Partially Paid"

image

donalnofrixion:
> If there's 0 matches on a filter type the tabbing doesn't work. In the below example the last tab selected was "Partially Paid"
> 
> image

The PR is not quite ready yet 😀 I thought I didn’t add any reviewers

donalnofrixion:
Currently can be tested in StoryBook under `Functional/PaymentrequestDashboard`.

The following token can be used to test against sandbox. It contains lots of PaymentRequests in various states.

`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0b2tlbmlkIjoiMTZjOGZjM2YtMjc2MC00NjE2LWIzMWUtZTljMzQwOWNkMDA3In0.yO1KpQNcti6ffBgBiBydSy4BIYlwTbo6Nd1x91aZzTI`

src/components/ui/PaymentRequestDashboard/Tabs.css
```diff
@@ -0,0 +1,44 @@
...
+
+.TabsTrigger:focus {
+  position: relative;
+}
+
+.TabsContent {
+  background-color: white;
+  padding: 1.5rem;
+  height: 100%;
+}
```
Publicker, line 44: If Tailwind was used in the component, we shouldn't need these classes.

src/components/ui/PrimaryButton/PrimaryButton.tsx
```diff
@@ -0,0 +1,17 @@
+interface PrimaryButtonProps {
+  label: string;
+  onClick?: () => void;
+}
+
+const PrimaryButton = ({ label, onClick }: PrimaryButtonProps) => {
+  return (
+    <span
```
Publicker, line 8: If we are rendering a button I'd suggest to use the `<button>` tag.

Regarding the UI, if we add the `transition` class, the hover transition would be a bit smoother.

src/web-components/PaymentRequestDashboard/PaymentRequestDashboard.tsx
Publicker, line 1: We use to add the `Web` prefix in these kind of files.

src/components/ui/PaymentRequestDashboard/Tabs.css
```diff
@@ -0,0 +1,44 @@
...
+
+.TabsTrigger[data-state='active'] {
+  color: #00264d;
+  background-color: white;
+  box-shadow: inset 0 2px 0 0 #00b2b2;
+  cursor: default;
+}
+
+.TabsTrigger[data-disabled] {
+  pointer-events: none;
```
Publicker, line 33: I'll also add `cursor: not-allowed;` so you'd get a nice OS mouse distinction.

Publicker: Here's the Tailwind documentation for it so you can see it: https://tailwindcss.com/docs/cursor

pabloi-nf (changes requested):
Thanks Donal, looking good.

About the top row:
- The main title "Payment requests needs to have only 16px left margin from the tab. Now has 48px.
- The main greenish button should have a right margin of 24px.
- The "settings" button should also have a background like the other one. On the default state should be transparent, and on hover should be #EDF2F7.

image


About the tabs:
- They need to have a different background color on the hover state. This color is #F0F2F5.
- This is not on the Figma, but I think the border top color of each tab can match the little dot color. That is: gray for the unpaid, orange for the partial and green for the paid.

Thanks

donalnofrixion:
Hi @pabloi-nf I've made the changes you requested.

pabloi-nf (changes requested):
Thanks Donal, it looks great.
There's one thing that I missed. The font-weight of the "settings" button is set at 500. I know is the weight in the Figma file, but it looks much more bold on the HTML.
Can we change it to 400? After that it's ready.

Thank you very much

pabloi-nf (approved):

PR #32, opened 2023-05-15
Added QR code component

sauravmaiti22:
Added QR code component


donalnofrixion:
The QR code should be able to be downloaded as per the Figma.

src/components/ui/QRCode/QRCode.tsx
```diff
@@ -0,0 +1,11 @@
+import QRCodeComponent from 'react-qr-code';
```
donalnofrixion, line 1: The QR code should be able to be downloaded as per the Figma.

donalnofrixion (changes requested):

sauravmaiti22:
> The QR code should be able to be downloaded as per the Figma.

Didn't see that in edit mode. Will add it now.

pabloi-nf (changes requested):
Hey @sauravmaiti22, I could download the QR, but it seems to be the same image we can see on the HTML. The idea for this would be to download a bigger image that can be used in other contexts, like printing and so on. 

I think at least we would need an image at 1024x1024px.

src/components/ui/QRCode/QRCode.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { AnimatePresence, motion } from 'framer-motion';
+
+const QRCode = ({ url }: { url: string }) => {
+  const [isHovered, setIsHovered] = useState(false);
+  const qrCodeRef = useRef<HTMLDivElement>(null);
+
+  const downloadAsImage = () => {
+    const element = qrCodeRef.current as HTMLDivElement;
+
+    html2canvas(element).then((canvas) => {
```
donalnofrixion, line 13: We should think about giving the downloaded QR file a meaningful name. Something that tells you what it is. How much, etc

donalnofrixion (changes requested):

src/components/ui/QRCode/QRCode.tsx
sauravmaiti22: Yes. I thought about it but it was getting really long. One suggestion would be nf-PR-{PaymentReqestID}. Adding amount in the name would be tricky with decimal.

donalnofrixion: It would be worth ask Andres and Pablo what they think.

sauravmaiti22: Updated PR. Changed the file name format to QR-DD-MM-YYYY-HHMMSS.

sauravmaiti22:
> Hey @sauravmaiti22, I could download the QR, but it seems to be the same image we can see on the HTML. The idea for this would be to download a bigger image that can be used in other contexts, like printing and so on.
> 
> I think at least we would need an image at 1024x1024px.



> Hey @sauravmaiti22, I could download the QR, but it seems to be the same image we can see on the HTML. The idea for this would be to download a bigger image that can be used in other contexts, like printing and so on.
> 
> I think at least we would need an image at 1024x1024px.

@pabloi-nf I have made the change to download the QR code in a bigger size.

donalnofrixion (approved):

Publicker (approved):

pabloi-nf (approved):

PR #34, opened 2023-05-16
Tag component

donalnofrixion:


arifmatin (approved):

src/components/ui/Tag/Tag.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <path d="M1 0.5L9 8.5" />
+        <path d="M9 0.5L1 8.5" />
+      </svg>
+    </div>
+  );
+};
+
+Tag.componentProps = {
+  label: String,
+};
```
Publicker, line 71: Not need as this is not gonna be a WebComponent

src/components/ui/Tag/Tag.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onDelete?: (id: string) => void;
+}
+
+const Tag = ({ id, label, onDelete }: TagProps) => {
+  const [text, setText] = useState(label);
+  const [deleteMode, setDeleteMode] = useState(false);
+
+  useEffect(() => {
+    deleteMode ? setText('Delete?') : setText(label);
+  }, [deleteMode]);
```
Publicker, line 17: During my testing, I noticed that the component didn't update when the label prop was changed.

I suspect that's due to the internal `text` state, which isn't necessarily bad. But we could improve it a bit by eliminating the text state and the useEffect. It could be even cleaner!

We can do this by doing something similar to the code below:

```tsx

const [deleteMode, setDeleteMode] = useState(false);
const text = !deleteMode ? label : "Delete?";

```

Publicker (commented):
What do you think about combining the `Tag` and `Chip` components? If we take a look at the Figma (image attached), it seems like they could just be variations of the same component, maybe with different sizes and without the delete function. I'm certainly open to thoughts on this, I just think it's a bit better to consolidate them in this instance since the differences are minor and can't see a reason to maintain them separately.

Figma - Tag component

pabloi-nf (approved):

Publicker:
> > What do you think about combining the `Tag` and `Chip` components? If we take a look at the Figma (image attached), it seems like they could just be variations of the same component, maybe with different sizes and without the delete function. I'm certainly open to thoughts on this, I just think it's a bit better to consolidate them in this instance since the differences are minor and can't see a reason to maintain them separately.
> > Figma - Tag component
> 
> I considered that but TBH I'm not sure it's worth doing.

Okay

Publicker (approved):

PR #35, opened 2023-05-17
Copy link component

sauravmaiti22:
Copy link component


arifmatin:
The Url link doesn't wrap when using a small screen.

copylibk

donalnofrixion:
In the Figma the 'Link copied' disappears after a few seconds to reveal the 'Copy link'

sauravmaiti22:
> In the Figma the 'Link copied' disappears after a few seconds to reveal the 'Copy link'

Done.

sauravmaiti22:
> The Url link doesn't wrap when using a small screen.
> 
> copylibk

I discussed this with Pablo I and we are now showing only part of the url if it's too big.

arifmatin (approved):

src/components/ui/CopyLink/CopyLink.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }, [copied]);
+
+  const copyLink = () => {
+    navigator.clipboard.writeText(link);
+    setCopied(true);
+  };
+
+  return (
+    <div className="flex items-center gap-[10px] h-10 rounded-full justify-between px-2 pr-4 shadow-[0_0_8px_rgba(4,41,49,0.15)] text-[13px] relative">
```
Publicker, line 23: Tailwind already has a 10px gap! It's under the name 'gap-2.5'. Check out the documentation right here: https://tailwindcss.com/docs/gap.

----

Whenever the value is not in Tailwind, as for the 13px, we always try to keep it in `rem` unit of measure. To transform it from `px` to `rem` I'd recommend you to use https://nekocalc.com/px-to-rem-converter as it's quite simple to use.

Whenever you can't find a specific value in Tailwind, like 13px for instance, we always try to convert it into the rem unit. A tool I'd recommend is https://nekocalc.com/px-to-rem-converter. It's quite easy to use.

(Please look throughout the file as this is present in different spots!)

---

Looking at the link, it seems a bit close to the left side. I reckon the padding in the Figma design was set to 16px, not 8px.

pabloi-nf (approved):

src/components/ui/CopyLink/CopyLink.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <div className="whitespace-nowrap overflow-x-clip">{link}</div>
+      <div className="w-14 h-6 bg-gradient-to-l from-white right-[88px] absolute"></div>
```
Publicker, line 25: Love this for the gradient! Good job here!

Just a very minimal comment here, which is just a slight improvement, as this element is actually only visible and not intractable for the user, I'd recommend adding the `pointer-events-none` class (docs [here](https://tailwindcss.com/docs/pointer-events)).

Publicker (changes requested):

src/components/ui/CopyLink/CopyLink.tsx
sauravmaiti22: I think you're right. I mistakenly changed it back to 8px because the right end padding has to be 8px. So I thought it would be consistent.

src/components/ui/CopyLink/CopyLink.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, [copied]);
+
+  const copyLink = () => {
+    navigator.clipboard.writeText(link);
+    setCopied(true);
+  };
+
+  return (
+    <div className="flex items-center gap-[10px] h-10 rounded-full justify-between px-2 pr-4 shadow-[0_0_8px_rgba(4,41,49,0.15)] text-[13px] relative">
+      <div className="whitespace-nowrap overflow-x-clip">{link}</div>
```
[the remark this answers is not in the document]
sauravmaiti22: Why is it not a good idea to user <div>?

src/components/ui/CopyLink/CopyLink.tsx
```diff
@@ -0,0 +1,75 @@
...
+                stroke-linecap="round"
+                stroke-linejoin="round"
+              />
+              <path
+                d="M9.87855 3.71145L10.3786 3.21145C10.9412 2.64884 11.7042 2.33276 12.4999 2.33276C13.2955 2.33276 14.0586 2.64884 14.6212 3.21145C15.1838 3.77406 15.4999 4.53713 15.4999 5.33278C15.4999 6.12844 15.1838 6.8915 14.6212 7.45412L11.4392 10.6361C10.9056 11.1706 10.1901 11.4841 9.43546 11.5142C8.68083 11.5442 7.94263 11.2885 7.36826 10.7981C6.79389 10.3077 6.4256 9.61878 6.33692 8.86877C6.24825 8.11877 6.44571 7.36291 6.88988 6.75212"
+                stroke="#454D54"
+                stroke-linecap="round"
+                stroke-linejoin="round"
+              />
+            </svg>
```
Publicker, line 49: I just saw Aaron raising a [ticket regarding this](https://nofrixion.atlassian.net/browse/MOOV-1764). It seems we're having some trouble when we put SVGs directly into the code. We need to make sure we're capitalising the dashed properties.

Here's what I mean:

`stroke-linecap` should be `strokeLinecap`
`stroke-linejoin` should be `strokeLinejoin`

---

This is a peculiar React quirk, so we should make sure we're doing this consistently across all SVG files. Since you're on it, could you please make these changes in the SVGs that have this issue? I believe there are only a few.

This only applies for SVGs being used directly on the code, not when they're being used from the `` tag.

Publicker (changes requested):

src/components/ui/CopyLink/CopyLink.tsx
sauravmaiti22: @Publicker I tried this and it changed the pointer to cursor: text. Without this, it was just a cursor: default. It's not part of the copy link button so it will not inherit the cursor.

src/components/ui/CopyLink/CopyLink.tsx
sauravmaiti22: Oh yes. I'm surprised that this sneaked in because I was getting error for the other svgs so I changed it. But didn't get it for this one. :(

sauravmaiti22: Done.

Publicker (approved):

src/components/ui/CopyLink/CopyLink.tsx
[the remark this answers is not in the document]
Publicker: Just for semantic reasons. It's not gonna strongly affect anything, but just good as for semantic standards.

https://www.w3schools.com/html/html5_semantic_elements.asp

PR #38, opened 2023-05-18
Amount paid component

sauravmaiti22:
For the progress bar I had to use style instead of className because tailwind flex-grow-[] does not take decimal value.

### DIFF
--- a/src/components/ui/AmountPaid/AmountPaid.tsx
+++ b/src/components/ui/AmountPaid/AmountPaid.tsx
@@ -7,24 +7,31 @@ const AmountPaid = ({
 }: {
   amountPaid: number;
   totalAmount: number;
-  currency: Currency;
+  currency: Currency.EUR | Currency.GBP;
 }) => {
-  const percentagePaid = (amountPaid / totalAmount) * 100;
+  const percentagePaid = amountPaid > totalAmount ? 100 : (amountPaid / totalAmount) * 100;
+  const outstandingAmount = amountPaid > totalAmount ? 0 : totalAmount - amountPaid;
   return (
     <>
       <div className="flex flex-col">
-        <span className="text-2xl font-semibold leading-8 mb-4 text-[#00264D]">
-          {currency == 'GBP' ? '£' : '€'}
-          {amountPaid.toFixed(2)}
+        <span
+          className={`text-2xl tabular-nums font-semibold leading-8 mb-4  ${
+            amountPaid === 0 ? 'text-[#73808C]' : 'text-[#00264D]'
+          }`}
+        >
+          {currency == Currency.GBP ? '£' : '€'}
+          {amountPaid === 0 ? totalAmount : amountPaid.toFixed(2)}
         </span>
         <div className="flex flex-row h-[0.125rem]">
           <div className="bg-[#40BFBF]" style={{ flexGrow: `${percentagePaid}` }}></div>
           <div className="bg-[#EDF2F7]" style={{ flexGrow: `${100 - percentagePaid}` }}></div>
         </div>
-        <span className="text-sm leading-5 text-[#73808C] font-medium mt-2">
-          {currency == 'GBP' ? '£' : '€'}
-          {(totalAmount - amountPaid).toFixed(2)} outstanding
-        </span>
+        {amountPaid > 0 && (
+          <span className="text-sm leading-5 text-[#73808C] font-medium mt-2">
+            {currency == Currency.GBP ? '£' : '€'}
+            <span className="tabular-nums">{outstandingAmount.toFixed(2)}</span> outstanding
+          </span>
+        )}
       </div>
     </>
   );
--- a/src/components/ui/AmountPaid/AmountPaid.tsx
+++ b/src/components/ui/AmountPaid/AmountPaid.tsx
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import { Currency } from '../../../api/types/Enums';
 
 const AmountPaid = ({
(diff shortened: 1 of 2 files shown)


Publicker (changes requested):
Everything seems to be in order, with just a couple of tiny tweaks I think.

In the Figma design you can see that the numbers display "financial figures" and, despite using the same font, in the Storybook we can see that the numbers appear different. This is due to something called "font features". For the Inter font, these features are 'cv01', 'cv02', 'cv03', 'cv04'. We previously implemented this with CSS. However, Tailwind now offers this feature (refer to this post: https://tailwindcss.com/blog/tailwindcss-v3-3#configure-font-variation-settings-for-custom-font-families), so you can set it up there!

> If you want to know more about this font features for the Inter font family, you can see kind of the "documentation" over here: https://rsms.me/inter/

Apart from this, I noticed an issue when the paid amount exceeds the total amount. The line doesn't seem to respond correctly. My suggestion would be to keep the line at full length in these cases, as it's plausible that an overpayment could occur with a payment request.

pabloi-nf (changes requested):
- The amount font-size needs to be 32px.
- Check the Inter font we are using. It seems to be different than the one using in the Figma (from Google Fonts).
- The line representing the percentage paid needs to have rounded edges.
- If there is an overpaid, the outstanding amount has to be 0. 
image

sauravmaiti22:
> Everything seems to be in order, with just a couple of tiny tweaks I think.
> 
> In the Figma design you can see that the numbers display "financial figures" and, despite using the same font, in the Storybook we can see that the numbers appear different. This is due to something called "font features". For the Inter font, these features are 'cv01', 'cv02', 'cv03', 'cv04'. We previously implemented this with CSS. However, Tailwind now offers this feature (refer to this post: https://tailwindcss.com/blog/tailwindcss-v3-3#configure-font-variation-settings-for-custom-font-families), so you can set it up there!
> 
> > If you want to know more about this font features for the Inter font family, you can see kind of the "documentation" over here: https://rsms.me/inter/
> 
> Apart from this, I noticed an issue when the paid amount exceeds the total amount. The line doesn't seem to respond correctly. My suggestion would be to keep the line at full length in these cases, as it's plausible that an overpayment could occur with a payment request.

Thanks @Publicker. This was very useful. I did notice the difference in font but I thought I was hallucinating 😄 . I have made the changes. Please have a look.

Publicker:
> > Everything seems to be in order, with just a couple of tiny tweaks I think.
> > In the Figma design you can see that the numbers display "financial figures" and, despite using the same font, in the Storybook we can see that the numbers appear different. This is due to something called "font features". For the Inter font, these features are 'cv01', 'cv02', 'cv03', 'cv04'. We previously implemented this with CSS. However, Tailwind now offers this feature (refer to this post: https://tailwindcss.com/blog/tailwindcss-v3-3#configure-font-variation-settings-for-custom-font-families), so you can set it up there!
> > > If you want to know more about this font features for the Inter font family, you can see kind of the "documentation" over here: https://rsms.me/inter/
> > 
> > 
> > Apart from this, I noticed an issue when the paid amount exceeds the total amount. The line doesn't seem to respond correctly. My suggestion would be to keep the line at full length in these cases, as it's plausible that an overpayment could occur with a payment request.
> 
> Thanks @Publicker. This was very useful. I did notice the difference in font but I thought I was hallucinating 😄 . I have made the changes. Please have a look.

@sauravmaiti22 Actually, disregard the font comment for now because something odd's happening there. Apparently the font works fine locally but it doesn't in remote, @pabloi-nf and myself are investigating.

sauravmaiti22:
> * The amount font-size needs to be 32px.
> * Check the Inter font we are using. It seems to be different than the one using in the Figma (from Google Fonts).
> * The line representing the percentage paid needs to have rounded edges.
> * If there is an overpaid, the outstanding amount has to be 0.
> 
> image

Yes 

> 

I have made the changes. Thanks @pabloi-nf

src/components/ui/AmountPaid/AmountPaid.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <div className="bg-[#EDF2F7] rounded-r-sm" style={{ flexGrow: `${100 - percentagePaid}` }}></div>
+        </div>
+        {amountPaid > 0 && (
+          <span className="text-sm leading-[1.063rem] text-[#73808C] font-medium mt-2">
+            {currency == Currency.GBP ? '£' : '€'}
+            <span className="tabular-nums">{outstandingAmount.toFixed(2)}</span> outstanding
+          </span>
+        )}
+      </div>
+    </>
```
Publicker, line 36: Just noticed a few more things.

1. Fragment (`<> </>`) is not needed as component returns only one element.
2. Substitution of classes should be done by using the `classNames` modules. [Here](https://github.com/nofrixion/nofrixion.webcomponents/blob/develop/src/components/ui/CreatePaymentRequestPage/CreatePaymentRequestPage.tsx#L409-L411)'s an example.
3. All HEX colours in here are actually available through names as they were added in the tailwind.config file.

Here are a few examples of classes to change:
- From `text-[#73808C]` to `text-greyText`
- From `text-[#73808C]` to `text-defaultText`
- From `bg-[#40BFBF]` to `bg-positiveActionBackground`
- and so on...

Publicker (changes requested):

src/components/ui/AmountPaid/AmountPaid.tsx
sauravmaiti22: > Just noticed a few more things.
> 
> 1. Fragment (`<> </>`) is not needed as component returns only one element.
> 2. Substitution of classes should be done by using the `classNames` modules. [Here](https://github.com/nofrixion/nofrixion.webcomponents/blob/develop/src/components/ui/CreatePaymentRequestPage/CreatePaymentRequestPage.tsx#L409-L411)'s an example.
> 3. All HEX colours in here are actually available through names as they were added in the tailwind.config file.
> 
> Here are a few examples of classes to change:
> 
> * From `text-[#73808C]` to `text-greyText`
> * From `text-[#73808C]` to `text-defaultText`
> * From `bg-[#40BFBF]` to `bg-positiveActionBackground`
> * and so on...

Oh thanks @Publicker. Will make these changes.

Publicker (approved):

Publicker (approved):

pabloi-nf (approved):
Besides the font issue everything else looks good

PR #43, opened 2023-05-22
Moov 1730 Tag Manager

donalnofrixion:


arifmatin (approved):
Really cool animation when adding the tags.

donalnofrixion:
https://github.com/nofrixion/nofrixion.webcomponents/assets/89344328/9408fcc2-e53c-4635-a7f4-6fa91d462cd8

Publicker (approved):
Hey @donalnofrixion, LGTM! Still, I have a few points I'd like to bring up.

- The `TagManager`'s add functionality is making me think a bit. I'm trying to think a step ahead about how it will mesh with the API work. There are a couple of things here. Firstly, the client side is creating the ID. While that's not necessarily bad, it's not common for us. Secondly, the current setup might make it tricky for us to catch errors. Even though our error handling could use some improvement, at least we can catch when something goes wrong. With the new setup, I don't see how we could.
- I reckon the Tag Manager could use another prop. Maybe we can call it `tagsToSuggest` or something similar. Right now, it's using the same tags that are being passed through props, which might limit the suggestions. For instance, if a PR only has one tag, and the merchant has more tags on their tags table, we should still suggest those.
- Lastly, have you considered adding an animation when a tag is deleted? Don't sweat it if it's too time-consuming, but it would be great if we can do it with minimal changes.

pabloi-nf (changes requested):
- When a tag is already added on this PR, it should appear in the list as disabled with a text color of C7CCD1. Otherwise I can think I'm adding the tag and nothing happens, which looks like an error. See video.

https://github.com/nofrixion/nofrixion.webcomponents/assets/127946572/cb57c898-9927-4e83-bc2b-6552da2645c5

- The animation when a tag is deleted should be a little slower. The "add tag" button moves too fast to the new position.

- When I start an interaction with a tag (to delete or add a new one) the status should reset if I click outside. Otherwise it can create strange situations like having an almost deleted tag while I'm trying to add a new one. See video.

https://github.com/nofrixion/nofrixion.webcomponents/assets/127946572/fc12fcf2-2d2a-4725-ae02-6d291cb299d1

- Pressing Escape on the keyboard should cancel the delete or add tag interaction too.

pabloi-nf (changes requested):
I'm noticing something odd that I didn't see before here. The row of tags looks OK, when in the tag manager section they look a little taller:
image

donalnofrixion:
Hi @pabloi-nf I've implemented the changes you requested.

1. Support escape key to exit edit or delete mode
2. Support click outside to exit edit or delete mode
3. Animation improvements

pabloi-nf (changes requested):
Hi Donal, 

I apologize if I wasn't clear in my previous comment. What I meant when I said to put the text in gray if a tag was already added was for the list, not for a duplicated tag.
The idea is to still show the tag name on the list, but to disable it, to avoid adding it for a second time:
image

donalnofrixion:
> Hi Donal,
> 
> I apologize if I wasn't clear in my previous comment. What I meant when I said to put the text in gray if a tag was already added was for the list, not for a duplicated tag. The idea is to still show the tag name on the list, but to disable it, to avoid adding it for a second time: image

@pabloi-nf Apologies. That makes much more sense!

donalnofrixion:
https://github.com/nofrixion/nofrixion.webcomponents/assets/89344328/6f9b2d37-8be7-430c-b37f-02ffdda91713

Hi @pabloi-nf That change has been implemented. Items will be disabled in dropdown if they are already in the list

pabloi-nf (approved):

PR #45, opened 2023-05-22
[MOOV-1773]: Fixed extra API calls caused by reference comparison

agranillonf:
Metrics and PaymentRequests endpoints were being called twice. That was because the hooks that invoked them were being executed twice because of the Date parameters.

The Date parameters were reworked into number (milliseconds) parameters, and there was a mismatch between the endDateRange default value and the DateRange component default endDateRange value.

### DIFF
--- a/src/api/hooks/usePaymentRequestMetrics.ts
+++ b/src/api/hooks/usePaymentRequestMetrics.ts
@@ -6,8 +6,8 @@ export const usePaymentRequestMetrics = (
   apiUrl: string,
   authToken: string,
   merchantId: string,
-  fromDate?: Date,
-  toDate?: Date,
+  fromDateMs?: number,
+  toDateMs?: number,
 ) => {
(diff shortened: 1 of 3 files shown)


donalnofrixion (approved):

Publicker (approved):
LGTM! Just one question, do we still need to transform the date to `ms` even after doing the `endOfDay` transformation?

arifmatin (approved):

agranillonf:
Yes. I tested the endOfDay change without the Ms transformation and the useEffect from the usePaymentRequestMetrics hook was still fired twice, because even if the Date objects were the same, the references weren't. Switching them to a native type prevents that.

Publicker:
> Yes. I tested the endOfDay change without the Ms transformation and the useEffect from the usePaymentRequestMetrics hook was still fired twice, because even if the Date objects were the same, the references weren't. Switching them to a native type prevents that.

Makes sense, thanks for that @agranillonf!

PR #49, opened 2023-05-24
Payment request details modal UI

sauravmaiti22:
This is the first PR for the payment request details modal UI. Need to do more work on payment attempts on API to be able to make it functional. Made some changes to local payment request object.


src/components/ui/Contact/Contact.tsx
```diff
@@ -1,10 +1,40 @@
...
+    },
+  },
+  defaultVariants: {
+    size: 'small',
+  },
+});
+
+interface ContactProps extends LocalContact {
+  size: 'small' | 'large';
+}
```
Publicker, line 31: I think you could extract the size variant using a helper that comes with the library.

See more here: https://cva.style/docs/getting-started#variantprops

src/components/ui/PaymentRequestStatusBadge/PaymentRequestStatusBadge.tsx
```diff
@@ -4,40 +4,60 @@ import partiallyPaidIcon from '../../../assets/images/paymentrequest_status_part
...
+    },
+  },
+  defaultVariants: {
+    size: 'small',
   },
 });
 
 interface PaymentRequestStatusBadgeProps {
   status: LocalPaymentStatus;
+  size: 'small' | 'large';
```
Publicker, line 40: Same about getting the type using the helper: https://cva.style/docs/getting-started#variantprops

src/components/ui/Transactions/Transactions.tsx
```diff
@@ -33,38 +25,40 @@ const Transactions = ({
...
+                {transaction.paymentMethod === LocalPaymentMethodTypes.Card && transaction.last4DigitsOfCardNumber && (
                   <div className="text-sm ml-1 items-center flex">
                     <span className="text-[0.375rem] mr-1">&#8226;&#8226;&#8226;&#8226;</span>
                     <span>{transaction.last4DigitsOfCardNumber}</span>
                   </div>
                 )}
               </div>
             </td>
-            <td className="pl-6 pb-2 pt-2 leading-6">
+            <td className={classNames('pl-6 pb-2 leading-6', { 'pt-2': index !== 0 })}>
```
Publicker, line 61: You might want to try if you'd achieve the same behaviour of the padding but without using `classNames` helper with just using the `gap-y-2` in the parent (`<tr>` element). It would look a bit cleaner and remove a bit of boilerplate.

It's fine to leave it if gap doesn't work (docs [here](https://tailwindcss.com/docs/gap))

src/components/ui/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+  paymentRequest,
+  onRefundClick,
+  onTagAdded,
+}: {
+  paymentRequest: LocalPaymentRequest;
+  onRefundClick: (paymentAttemptID: string) => void;
+  onTagAdded: (tag: Tag) => void;
+}) => {
+  return (
+    <>
```
Publicker, line 22: Fragment could be removed

src/utils/mockedData.ts
```diff
@@ -39,48 +124,55 @@ const regular: PaymentRequest = {
...
+  paymentAttempts: mockPaymentAttempts,
 };
 
-const noShippingAddress: PaymentRequest = {
+const partiallyPaidPaymentRequest = (): LocalPaymentRequest => {
+  const partiallyPaid: LocalPaymentRequest = { ...regular };
+  partiallyPaid.paymentAttempts = partiallyPaidMockPaymentAttempts;
+  partiallyPaid.status = 'partial';
+  return partiallyPaid;
+};
```
Publicker, line 146: You can get the same output without the need for a function.

Consider this plain old object as an example. You can easily adapt it to your situation:

```jsx
const paymentRequest = {
  id: 1,
  status: "paid",
  name: "🤘🏻",
}
```

If you want to change the get the same object but with a different status in a different variable, you'd do the following:
```jsx
const unpaidPaymentRequest = {
  ...paymentRequest, // This is your base object
  status: "unpaid", // This is the property you want to switch up
  // Fancy making more changes? Just add them here
}
```

Publicker (changes requested):
First up, fantastic job Saurav! The progress is impressive.

I've a couple of broader comments, apart from the minor ones already shared:

- The current PR lacks the incorporation of modal behaviour, despite the component being named 'Modal.' For instance, if you look at the `UI/Payment Conditions Modal` or `UI/Payment Methods Modal` components, they contain modal logic. That is, they accept an 'open' property which, when modified, controls the opening or closing of the modal.
- Once you have the modal behaviour sorted, it would be a great idea to integrate it with the `UI/Payment Request Table component`. With this integration, clicking on a row will trigger the modal, and the modal will close when the user decides to do so.

Including these changes in the current PR makes sense, as they are UI-related and will streamline the development of the functional component

src/utils/mockedData.ts
sauravmaiti22: Wow! This is so much easier. Thanks Pablo!

pabloi-nf (changes requested):
- Top margin here needs to be 24px: 

image



- The line doesn't seem to match with the percentage completed:
image



- The Add Tag text looks darker than it should be. In Donal's PR looks OK:

image

src/components/ui/Transactions/Transactions.tsx
sauravmaiti22: I tried this and it didn't work unfortunately 😞

src/components/ui/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
Publicker: Okay, no worries

src/components/ui/Transactions/Transactions.tsx
Publicker: Okay, no worries

src/components/ui/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
Publicker: Okay, disregard that comment then.

sauravmaiti22:
> First up, fantastic job Saurav! The progress is impressive.
> 
> I've a couple of broader comments, apart from the minor ones already shared:
> 
> * The current PR lacks the incorporation of modal behaviour, despite the component being named 'Modal.' For instance, if you look at the `UI/Payment Conditions Modal` or `UI/Payment Methods Modal` components, they contain modal logic. That is, they accept an 'open' property which, when modified, controls the opening or closing of the modal.
> * Once you have the modal behaviour sorted, it would be a great idea to integrate it with the `UI/Payment Request Table component`. With this integration, clicking on a row will trigger the modal, and the modal will close when the user decides to do so.
> 
> Including these changes in the current PR makes sense, as they are UI-related and will streamline the development of the functional component

@Publicker Yeah I thought about doing it in the next PR. Wanted to get it reviewed in pieces and not make it too complicated for reviewers. About the modal bit, I think I will call this just PaymentRequestDetails and create a separate modal component as I don't want to clutter this one. I kind of want to control the modal from the parent by wrapping this component.

src/utils/mockedData.ts
```diff
@@ -1,34 +1,119 @@
...
+export const mockPaymentAttempts: LocalPaymentAttempt[] = [
+  {
+    paymentAttemptID: 'a3b752d2-c0a6-4846-90e5-d783bb4ec005',
+    occurredAt: new Date('2023-05-18'),
+    paymentMethod: LocalPaymentMethodTypes.Card,
+    amount: 20.02,
+    currency: Currency.EUR,
+    processor: 'Visa',
+    last4DigitsOfCardNumber: '1234',
+  },
```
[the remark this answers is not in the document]
sauravmaiti22: I think that will be incorporated in the tag manager component if I'm not wrong. Because the Add Tag component does not display the tags.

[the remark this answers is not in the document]
Publicker: Ahh, okay - Then maybe just add a comment in the `PaymentRequestDetails` component to change the `AddTag` component to the `TagManager` component.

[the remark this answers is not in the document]
sauravmaiti22: Yes, I had added the comment.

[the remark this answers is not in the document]
Publicker: 🤦🏼‍♂️ You're right - apologies.

src/components/ui/PaymentRequestStatusBadge/PaymentRequestStatusBadge.tsx
sauravmaiti22: Done.

src/utils/mockedData.ts
sauravmaiti22: Done.

pabloi-nf:
@sauravmaiti22 the big number should represent the total amount to be paid. This was what we discussed on the meeting, with the idea that this number remains constant always. If that's the case the bar should be almost empty, since only 1.18 euros were paid.

For the tag, I'm talking about the color, not the weight. It should be the gray text color (#73808C).

Publicker (approved):

sauravmaiti22:
> @sauravmaiti22 the big number should represent the total amount to be paid. This was what we discussed on the meeting, with the idea that this number remains constant always. If that's the case the bar should be almost empty, since only 1.18 euros were paid.
> 
> For the tag, I'm talking about the color, not the weight. It should be the gray text color (#73808C).
@pabloi-nf Done.

donalnofrixion:
There's an issue with the 'Add tag' dropdown's opacity. It is also difficult to click it.
image

sauravmaiti22:
> First up, fantastic job Saurav! The progress is impressive.
> 
> I've a couple of broader comments, apart from the minor ones already shared:
> 
> * The current PR lacks the incorporation of modal behaviour, despite the component being named 'Modal.' For instance, if you look at the `UI/Payment Conditions Modal` or `UI/Payment Methods Modal` components, they contain modal logic. That is, they accept an 'open' property which, when modified, controls the opening or closing of the modal.
> * Once you have the modal behaviour sorted, it would be a great idea to integrate it with the `UI/Payment Request Table component`. With this integration, clicking on a row will trigger the modal, and the modal will close when the user decides to do so.
> 
> Including these changes in the current PR makes sense, as they are UI-related and will streamline the development of th

> There's an issue with the 'Add tag' dropdown's opacity. It is also difficult to click it. image

@donalnofrixion Let me check.

pabloi-nf (approved):

donalnofrixion (approved):

PR #50, opened 2023-05-24
Implement New Status Filter Design for Payment Requests (ex Tabs)

Publicker:
This pull request introduces a new design for the Status filter (previously referred to as Tabs) in the Payment Requests Table.

Notably, accessibility features have been maintained in this implementation. Users can navigate to the status filter using the TAB key, and then switch between statuses using the arrow keys (← and →). Importantly, the status will not change when continuing to press the TAB key.

Demo video:

https://github.com/nofrixion/nofrixion.webcomponents/assets/52673485/8ef117c8-d0e6-48f5-ab15-27cf75d8a702

### DIFF
--- a/src/api/hooks/usePaymentRequestMetrics.ts
+++ b/src/api/hooks/usePaymentRequestMetrics.ts
@@ -11,9 +11,11 @@ export const usePaymentRequestMetrics = (
 ) => {
   const [metrics, setMetrics] = useState<PaymentRequestMetrics>();
   const [apiError, setApiError] = useState<ApiError>();
+  const [isLoading, setIsLoading] = useState<boolean>(false);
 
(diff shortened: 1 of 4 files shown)


donalnofrixion (approved):

pabloi-nf (approved):

Publicker:
@pabloi-nf I just realised I hadn't included the loading state for the tabs. It's all set now! Could you run the same network throttling tests you used on the other skeletons?

Also, we've got a new Storybook parameter called isLoading for the UI/Tab component to play around with it to see how the loading state looks.

Note: you might see the tabs loader disappear a bit faster than the table loader. This is because the tabs' data comes from a different API call than the table rows.

PR #51, opened 2023-05-25
Implement Empty and Initial State for Payment Requests

Publicker:
This pull request introduces the empty and initial state for the Payment Requests table. These states have been designed to provide feedback to the user when no payment requests are found based on filters or when no payment requests have been created.

Changes include:

- Adding an empty state for instances where no payment requests are found that match the selected filters. Currently, the only filter available is by status.
- Implementing an initial state for scenarios where no payment requests have been created. This is likely applicable for newly onboarded merchants.
- Introducing an animation to transition from the loading state to the initial state. As we cannot determine in advance whether the merchant has any Payment Requests, this transition aids to improve the UX. During this transition, the status button filters are hidden with an opacity change, and the table below moves upward. The 'Create Payment' button in the top-right also follows the same exit animation.

Two videos demonstrating these transitions under different network conditions (slow and regular internet) have been attached to this pull request.

Note to @pabloi-nf: You have been added to a merchant with no Payment Requests in the Dev environment. The ID for this merchant is `5A0A6425-C25D-4A9C-D8AF-08DB0F560F56`. Please remember to acquire a token for this environment and switch the `apiUrl` to the `dev` one.

Slow internet Video:

https://github.com/nofrixion/nofrixion.webcomponents/assets/52673485/cf7e4989-4246-4e4e-9c3d-627734d99b10

Regular internet Video:

https://github.com/nofrixion/nofrixion.webcomponents/assets/52673485/ccd83ae0-b32e-4dd3-a0fe-887984589bfe


pabloi-nf (changes requested):
- Date component shouldn't be on the initial state:

image

- We should use this text:

image

- We have to avoid the jump it makes when the image still not loaded.

donalnofrixion (approved):

src/components/functional/PaymentRequestDashboard/PaymentRequestDashboard.tsx
```diff
@@ -117,6 +119,8 @@ const PaymentRequestDashboard = ({
     await fetchPaymentRequests();
   };
 
+  console.log('LALA: ', isLoadingMetrics || (metrics && metrics?.all > 0));
```
sauravmaiti22, line 122: Can remove this.

Publicker:
> * Date component shouldn't be on the initial state:
> 
> image
> * We should use this text:
> 
> image
> * We have to avoid the jump it makes when the image still not loaded.

@pabloi-nf All three items are sorted out! Plus, the date component fades away too :)

pabloi-nf (approved):

PR #53, opened 2023-05-30
User payment defaults

donalnofrixion:


arifmatin (approved):

donalnofrixion:
Hi @Publicker I've made those changes we discussed yesterday.

src/components/functional/CreatePaymentRequestPage/CreatePaymentRequestPage.tsx
```diff
@@ -83,13 +86,48 @@ const CreatePaymentRequestPage = ({
...
+    const client = new ClientSettingsClient(apiUrl, token);
+    const response = await client.saveUserPaymentDefaults(userPaymentDefaults);
+
+    if (response.error) {
+      makeToast('error', response.error.title);
+      return;
+    }
+  };
+
+  const getDefaultUserPaymentDefaults = (): UserPaymentDefaults => {
```
Publicker, line 99: This could be rather just a variable, there's no need for a function. We'd even have this into a `constants` file.

src/components/ui/EditOptionCard/EditOptionCard.tsx
```diff
@@ -1,32 +1,55 @@
...
   label: string;
   value?: string;
   details?: string[];
   onClick?: () => void;
   children?: React.ReactNode;
+  isLoading: boolean;
 }
 
-const EditOptionCard = ({ label, value, details, onClick, children }: EditOptionCardProps) => {
+// TODO: Disable edit mode when isLoading is true
```
Publicker, line 15: This seems to be already implemented.

src/components/ui/EditOptionCard/EditOptionCard.tsx
Publicker, line 1: Could we add two Storybook stories, one for each `UI/Edit Option Card` that we have (that being `UI/Edit Option Card/Payment Conditions` and `UI/Edit Option Card/Payment Methods`) but in their respective loading state?

Publicker (changes requested):

PR #55, opened 2023-05-31
Functional payment request details modal.

sauravmaiti22:
- Made the payment request details modal functional.
- Getting payment attempts from the api.
- Made the tag manager functional.
- Refund functionality for card will be implemented in subsequent PR.
- Display of processor i.e MasterCard/Visa/Bank name will be done in MOOV-1807.
- Using contentType as json for all API requests now. Have tested it and it works.

- Issues right now - 1. When some payments are in authorized state, the PR status remains as Partially Paid and the amount outstanding is 0. It is confusing as their is not indication that some payments are in authorized state.
2. Cannot add tags after payment is made on the PR because the update PR api does not allow that. Need discussion about this.

### DIFF
--- a/src/components/functional/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
+++ b/src/components/functional/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
@@ -1,0 +1,127 @@
+import { useEffect, useState } from 'react';
+import { useMerchantTags } from '../../../api/hooks/useMerchantTags';
+import UIPaymentRequestDetailsModal from '../../ui/PaymentRequestDetailsModal/PaymentRequestDetailsModal';
+import { RemotePaymentRequestToLocalPaymentRequest, parseApiTagToLocalTag } from '../../../utils/parsers';
+import { LocalTag } from '../../../types/LocalTypes';
+import { PaymentRequestClient } from '../../../api/clients/PaymentRequestClient';
+import { PaymentRequest, PaymentRequestUpdate } from '../../../api/types/ApiResponses';
+import { MerchantClient } from '../../../api/clients/MerchantClient';
+
+interface PaymentRequestDetailsModalProps {
+  token: string; // Example: "eyJhbGciOiJIUz..."
+  apiUrl: string; // Example: "https://api.nofrixion.com/api/v1"
+  merchantId: string;
+  paymentRequestID: string;
+  open: boolean;
+  onDismiss: () => void;
+}
+const PaymentRequestDetailsModal = ({
+  token,
+  apiUrl,
+  merchantId,
+  paymentRequestID,
+  open,
+  onDismiss,
+}: PaymentRequestDetailsModalProps) => {
+  const paymentRequestClient = new PaymentRequestClient(apiUrl, token, merchantId);
+  const merchantClient = new MerchantClient(apiUrl, token, merchantId);
+  const merchantTags = useMerchantTags(token, merchantId, apiUrl);
+  const [localMerchantTags, setLocalMerchantTags] = useState<LocalTag[]>([] as LocalTag[]);
+  const [paymentRequest, setPaymentRequest] = useState<PaymentRequest | undefined>();
+
+  const onRefundClick = async (paymentAttemptID: string) => {
+    //TODO: Will implement refund for atleast card payment attempts. For PISP, it will need to be worked on later.
+    console.log(paymentAttemptID);
+  };
+
+  const onTagAdded = async (tag: LocalTag) => {
+    if (paymentRequest) {
+      const existingTagIds = paymentRequest.tags?.map((tag) => tag.id) ?? [];
+      const paymentRequestUpdate: PaymentRequestUpdate = {
+        tagIds: existingTagIds.concat(tag.id),
+      };
+      const paymentRequestTagAdd = await paymentRequestClient.update(paymentRequest.id, paymentRequestUpdate);
+      if (paymentRequestTagAdd.error) {
+        console.log(paymentRequestTagAdd.error);
+      } else {
+        setPaymentRequest(paymentRequestTagAdd.data);
+      }
+    }
+  };
+
+  const onTagCreated = async (tag: LocalTag) => {
+    if (paymentRequest) {
+      const response = await merchantClient.addTag(parseApiTagToLocalTag(tag));
+      if (response.error) {
+        console.log(response.error);
+      } else {
+        const createdTag = response.data;
+        if (createdTag) {
+          const existingTagIds = paymentRequest.tags?.map((tag) => tag.id) ?? [];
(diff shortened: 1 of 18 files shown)


Publicker (changes requested):
I've had a good look through the stories and a few things caught my eye.

- First off, when you launch the modal from the UI component (not from the Functional Dashboard), it seems to pop up already scrolled. So, I can't see the top. But from the Functional Dashboard, it works just fine.
- I also noticed that the modal doesn't really have an entrance animation (I was picturing something like a right-to-left slide), though it does have a decent exit one.
- Once the modal is open, it seems like the History transactions are the focus point (you can spot a blue border around them indicating that they've got the accessibility focus). We shouldn't lose the blue border as it's a key accessibility feature. However, it probably shouldn't be the first thing to grab the focus. Maybe it's best not to select anything as soon as the modal opens.
- Speaking of the history transactions, they don't appear to fill the entire width, it's more like half.
- The copy link isn't the same as the one from the table when you hit the ellipsis and then "Copy Payment Link" (it should be the same from the table since we recently fixed that one).
- Maybe we could consider adding the Payment Request Details modal when tapping on a row in the `UI/Payment Request Table` Storybook story?

There are likely a few more details, but those might be best hashed out in a discussion with @pabloi-nf. One thing that struck me was when the Payment Request lacks contact info (like name and email), the component feels a bit off, especially at the top where it's a bit empty. Anything we can do about that, @pabloi-nf? What do you think?

I get a similar vibe with the History tab when there aren't any transactions (when the PR hasn't been paid yet). Maybe we could add an illustration or just a plain message that there's no history because there haven't been any transactions or no one has paid yet? Over to you, @pabloi-nf.

src/components/ui/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+            enter="transform transition ease-in-out duration-300"
+            enterFrom="translate-x-full"
+            enterTo="translate-x-0"
+            leave="transform transition ease-in-out duration-300"
+            leaveFrom="translate-x-0"
+            leaveTo="translate-x-full"
+          >
+            <Dialog.Panel>
+              <div className="absolute left-0 top-0 -ml-8 flex pr-2 pt-4 sm:-ml-10 sm:pr-4">
+                <button type="button" className="opacity-0 focus:outline-none" onClick={onDismiss}>
```
Publicker, line 55: I'd recommend changing the `opacity-0` to `invisible` as we won't get the cursor-pointer when accidentally hovering over an invisible button

Publicker (changes requested):
From the code point is almost there, there're just two small comments.

src/components/ui/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
sauravmaiti22: invisible makes it hidden and the focus goes to the transactions table.

src/components/functional/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (merchantTags.tags) {
+      setLocalMerchantTags(merchantTags.tags.map((tag) => parseApiTagToLocalTag(tag)));
+    }
+  }, [merchantTags.tags]);
+
+  useEffect(() => {
```
donalnofrixion, line 96: There is a custom hook available for this. usePaymentRequest
Also why can't the payment request be passed in to this from the PaymentRequestRow? It seems like when the modal is opened it fetches the payment request when it doesn't need to. This causes a noticeable delay.

src/components/functional/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+      const paymentRequestTagAdd = await paymentRequestClient.update(paymentRequest.id, paymentRequestUpdate);
+      if (paymentRequestTagAdd.error) {
+        console.log(paymentRequestTagAdd.error);
+      } else {
+        setPaymentRequest(paymentRequestTagAdd.data);
+      }
+    }
+  };
+
+  const onTagCreated = async (tag: LocalTag) => {
```
donalnofrixion, line 52: When tags are modified the payment request row doesn't update. It think this can be solved if the payment request is passed in by the PaymentRequestRow component.

donalnofrixion (changes requested):

src/components/functional/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }, [selectedPaymentRequestID, paymentRequests]);
+
+  const onRefundClick = async (paymentAttemptID: string) => {
+    //TODO: Will implement refund for atleast card payment attempts. For PISP, it will need to be worked on later.
+    console.log(paymentAttemptID);
+  };
+
+  const updatePaymentRequests = (updatedPaymentRequest: PaymentRequest) => {
+    for (let i = 0; i < paymentRequests.length; i++) {
```
donalnofrixion, line 53: You can probably use paymentRequests.findIndex instead.

donalnofrixion (approved):

src/components/functional/PaymentRequestDetailsModal/PaymentRequestDetailsModal.tsx
sauravmaiti22: Thanks! This is nice!

Publicker (approved):

pabloi-nf (changes requested):
- The description is not being displayed when creating a PR from sandbox. Is this because is the old portal?
- The amount is not showing decimals.
- The tabular numbers look odd in the amount field. We should remove this feature in this case.

donalnofrixion:
Hey @sauravmaiti22 Just noticed on the Payment Info details the payment methods are not being correctly set.

image

This payment request (e93d98e2-47b9-4bd4-be02-b16b30d04236) has more than just cards set.

sauravmaiti22:
> Hey @sauravmaiti22 Just noticed on the Payment Info details the payment methods are not being correctly set.
> 
> image
> 
> This payment request (e93d98e2-47b9-4bd4-be02-b16b30d04236) has more than just cards set.

Okay let me have a look.

sauravmaiti22:
> * The description is not being displayed when creating a PR from sandbox. Is this because is the old portal?
> * The amount is not showing decimals.
> * The tabular numbers look odd in the amount field. We should remove this feature in this case.

This is done.

PR #56, opened 2023-06-01
moov 1794 payments notifications email modal

donalnofrixion:
Payments notifications modal


pabloi-nf (changes requested):
Donal, the implementation is perfect, but on a second look I'm thinking that it looks a little crowded with so much text. I think we should reduce the margin bottom of the paragraph, and change its text to: "Send a notification to the specified email addresses when the payment is completed."

Also, we should change the text under the filed to say "Use commas to include more than one email address."

Does that sounds better?

donalnofrixion:
> Donal, the implementation is perfect, but on a second look I'm thinking that it looks a little crowded with so much text. I think we should reduce the margin bottom of the paragraph, and change its text to: "Send a notification to the specified email addresses when the payment is completed."
> 
> Also, we should change the text under the filed to say "Use commas to include more than one email address."
> 
> Does that sounds better?

Thanks @pabloi-nf I've changed the text and reduced the bottom margin from 48px to 24px.
image

pabloi-nf (approved):

PR #59, opened 2023-06-02
[MOOV-1813]: Added scrollbar-gutter to root HTML element

agranillonf:
The Payment Request Dashboard was experiencing visual glitches when displaying empty sections due to dynamic scrollbar visibility.

Now, a scrollbar-gutter CSS class is added to the root HTML element to avoid that.


Publicker (commented):
Have you tried building this as Web Component and updating it on the Portal locally to check if this fix works?

Would be interesting to see if the web component has access to the `document` and apply this type of properties to it.

agranillonf:
@Publicker Can confirm that it works on the Portal. It correctly sets the style to the HTML element and the visual glitch is no longer present.

Publicker (approved):

Publicker:
> @Publicker Can confirm that it works on the Portal. It correctly sets the style to the HTML element and the visual glitch is no longer present.

Lovely, thanks!

PR #64, opened 2023-06-16
[MOOV-1814,MOOV-1870]: Payment request deletion bug fixes

agranillonf:
- Fixed Payment Request Details side panel showing when clicking either deletion button.
- Added disabled Payment Request Delete button state. Now, only payment requests with no recorded events will be able to be deleted.

### DIFF
--- a/src/api/types/ApiResponses.ts
+++ b/src/api/types/ApiResponses.ts
@@ -46,7 +46,6 @@ export type PaymentRequest = {
   cardStripePaymentIntentID?: string;
   cardStripePaymentIntentSecret?: string;
   addresses: PaymentRequestAddress[];
-  events: PaymentRequestEvent[];
   jwk?: string;
   tags: Tag[];
   priorityBankID?: string;
--- a/src/api/types/ApiResponses.ts
+++ b/src/api/types/ApiResponses.ts
@@ -46,6 +46,7 @@ export type PaymentRequest = {
   cardStripePaymentIntentID?: string;
   cardStripePaymentIntentSecret?: string;
   addresses: PaymentRequestAddress[];
+  events: PaymentRequestEvent[];
   jwk?: string;
   tags: Tag[];
   priorityBankID?: string;
--- a/src/components/ui/InfoTooltip/InfoTooltip.tsx
+++ b/src/components/ui/InfoTooltip/InfoTooltip.tsx
@@ -10,7 +10,7 @@ interface InfoTooltipProps {
   side?: 'top' | 'bottom' | 'left' | 'right';
 }
 
-const InfoTooltip = ({ content, children, side }: InfoTooltipProps) => {
+const InfoTooltip = ({ content, children, side = 'top' }: InfoTooltipProps) => {
   const [open, setOpen] = useState(false);
 
   return (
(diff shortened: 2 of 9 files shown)


src/types/LocalEnums.ts
```diff
@@ -13,6 +13,32 @@ export enum LocalAddressType {
   Billing = 'Billing',
 }
 
+export enum LocalPaymentRequestEventType {
```
donalnofrixion, line 16: The property paymentAttempts already exists. It could be used instead of the events.

src/components/ui/PaymentRequestActionMenu/PaymentRequestActionMenu.tsx
```diff
@@ -55,6 +57,7 @@ const PaymentRequestActionMenu = ({ onDuplicate, onCopyLink, onDelete, onBlur }:
   const onDuplicateClick = (e: React.MouseEvent<HTMLDivElement>) => handleClick(e, onDuplicate);
   const onCopyLinkClick = (e: React.MouseEvent<HTMLDivElement>) => handleClick(e, onCopyLink);
   const onDeleteClick = (e: React.MouseEvent<HTMLDivElement>) => handleClick(e, onDelete);
+  const emptyClick = (e: React.MouseEvent<HTMLDivElement>) => e.stopPropagation();
```
donalnofrixion, line 60: Not sure but the tailwind class 'pointer-events-none' might work instead?

donalnofrixion (changes requested):

src/components/ui/PaymentRequestActionMenu/PaymentRequestActionMenu.tsx
agranillonf: Tried it, but the click event still propagates.

donalnofrixion (approved):

sipsorcery (approved):

arifmatin (approved):

src/utils/parsers.ts
Publicker, line 1: Both `Currency` and `events` seem to not be in use.

src/components/ui/InfoTooltip/InfoTooltip.tsx
```diff
@@ -26,7 +27,7 @@ const InfoTooltip = ({ content, children }: InfoTooltipProps) => {
         <AnimatePresence>
           {open && (
             <Tooltip.Portal forceMount>
-              <Tooltip.Content sideOffset={5} side="top" asChild>
+              <Tooltip.Content sideOffset={5} side={side ?? 'top'} asChild>
```
Publicker, line 30: It might be easier to read if we declare "top" as default param when deconstructing the props.

So, the line 13 would be like:
```tsx
const InfoTooltip = ({ content, children, side = "top" }: InfoTooltipProps) => {
```

And then, this line would be like:

```suggestion
              <Tooltip.Content sideOffset={5} side={side} asChild>
```

src/api/types/ApiResponses.ts
```diff
@@ -46,6 +46,7 @@ export type PaymentRequest = {
   cardStripePaymentIntentID?: string;
   cardStripePaymentIntentSecret?: string;
   addresses: PaymentRequestAddress[];
+  events: PaymentRequestEvent[];
```
Publicker, line 49: It seems like you've added `events`. I presume you intended to use them to enable or disable the delete button, but you ended up using `attempts` instead. Was that intentional?

And while we're on the subject, what are the differences between them?

pabloi-nf (approved):

src/api/types/ApiResponses.ts
agranillonf: I originally intended to use `events`, but was informed that `attempts` was a better choice since it's practically the same as `events` and it was readily available to use in the code.

PR #71, opened 2023-06-23
Keep row highlighted when details modal is shown

donalnofrixion:
https://github.com/nofrixion/nofrixion.webcomponents/assets/89344328/464529e7-f11d-4bdc-aba2-392b33ba63df


sipsorcery (approved):

sipsorcery:
Very nice! I will use that every day.

PR #74, opened 2023-06-28
[MOOV-1864]: Added new Payment Request Dashboard filter controls

agranillonf:
Added the ability to filter the payment request dashboard by:

- Text: On Merchant's name, payment request description and title, customer email address and PISP reference.
- Currency.
- Minimum requested amount.
- Maximum requested amount.
- Tags

The API client was also updated to support these filters.


src/components/ui/PaymentRequestAmountFilter/PaymentRequestAmountFilter.tsx
```diff
@@ -0,0 +1,197 @@
+﻿import FilterButton from '../FilterButton/FilterButton';
```
donalnofrixion, line 1: I'd remove PaymentRequest from the name. We'll reuse these for Accounts Payable.

src/components/ui/PaymentRequestTagFilter/PaymentRequestTagFilter.tsx
```diff
@@ -0,0 +1,104 @@
+﻿import FilterButton from '../FilterButton/FilterButton';
```
donalnofrixion, line 1: Same here

src/components/ui/PaymentRequestFilterRow/PaymentRequestFilterRow.tsx
```diff
@@ -0,0 +1,56 @@
+﻿import DateRangePicker, { DateRange } from '../DateRangePicker/DateRangePicker';
```
donalnofrixion, line 1: Maybe here too?

donalnofrixion:
Looks awesome Axel! very cool. One small thing. Should the Tag filter have a clear 'X' similar to the amount filter?
image

donalnofrixion (changes requested):

donalnofrixion:
Selecting pounds in the amount filter doesn't seem to work.
image

donalnofrixion:
Should the tags be alphabetically ordered?
image

pabloi-nf (changes requested):
Tags
- Tag cloud should be left aligned.
- Tag component buttons are missing the hover state.
- If the user clicks on "cancel" the tags selected should go back to the last state before opening the box. Video attached showing the problem. In the video you can see that I turned on 2 tags, clicked cancel, but when the dropdown is open again the tags are still selected, even when I didn't press on "Apply".
- Clicking outside the box should be the same as clicking "Cancel".
- Font-weight of the cancel button should be 400.

https://github.com/nofrixion/nofrixion.webcomponents/assets/127946572/ba5c88d2-a8b2-4a43-a935-002104ec22e3



Amount:
- The same cancel issue happens here.
- I made a mistake in the figma. The icon should be always the circle with the euro inside, not the filter icon.
- I find out it looks confusing if we don't put a border around the selected filter. We should do something like this 
image


Search:
- We should add an icon to reset the field. This can be an X like we have in the other ones.
image


Thanks!

agranillonf:
> Looks awesome Axel! very cool. One small thing. Should the Tag filter have a clear 'X' similar to the amount filter? image

@donalnofrixion There's no close button on the Figma, so I didn't put it in.

agranillonf:
> Selecting pounds in the amount filter doesn't seem to work. image

@donalnofrixion It doesn't work because the necessary changes to the API haven't been deployed to develop as of this comment.

agranillonf:
All comments have been attended and the currency filter changes have been deployed to the DEV API.

donalnofrixion (approved):

Publicker (approved):

pabloi-nf (changes requested):
Everything looks great! The only thing I'm seeing that I didn't notice before is that numbers don't have commas to separate thousands and millions 
image

Publicker (approved):

pabloi-nf (approved):

Publicker:
🎉 🎉 🎉

donalnofrixion (approved):

PR #85, opened 2023-07-11
Update Responsiveness for PR Dashboard and PR Details

Publicker:
This pull request includes modifications to the Payment Request dashboard and Payment Request page to enhance responsiveness. Please note that there are a few known areas for improvement that will be addressed in subsequent PRs. 

These include:

- The position of the 'Confirm PR' button in Review mode is not fixed to the bottom when the content is smaller than what it's inside.
- The tag manager currently defaults to a vertical orientation in mobile view, similar to the desktop view.
- The UI for date and sort filters in mobile view requires updates.

Any issues not listed above should be considered for feedback and future fixes.

Please review the changes and provide any necessary feedback.

### DIFF
--- a/src/assets/icons/close-button-icon.svg
+++ b/src/assets/icons/close-button-icon.svg
@@ -1,0 +1,6 @@
+<svg width="24" height="22" viewBox="0 0 24 22" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g id="Close Button">
+<path id="Vector 11" d="M22 21L12 11L22 1" stroke="#ABB3BA" stroke-linecap="round" stroke-linejoin="round"/>
+<path id="Vector 15" d="M2 0.999999L12 11L2 21" stroke="#ABB3BA" stroke-linecap="round" stroke-linejoin="round"/>
+</g>
+</svg>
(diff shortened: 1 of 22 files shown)


donalnofrixion (approved):

agranillonf (approved):

arifmatin (approved):

pabloi-nf (changes requested):
- Title font size should be 1.5 rem.
- There shouldn't be a scroll bar on the tabs.
image
- Let's use 1.75rem for the tabs numbers font size.
- Let's change the contact name font weight to 400.
- The create payment request button should have 12px of padding top and bottom.
- The title and the content in the op ups is not aligned 
image
- After creating a payment request, I can't scroll anymore on the list (I don't know if this is a problem with Storybook).
- The 3 dots should look like they have the same distance with the edge of the card and with the EUR text 
image
- We should redesign the pager, it's too small to be used on mobile.
- I'm making new payment methods icons to adapt better to this size.

Publicker:
@pabloi-nf Most comment were addressed.

Those were:
- ✅ Title font size should be 1.5 rem.
- ✅ There shouldn't be a scroll bar on the tabs.
- ✅ Let's use 1.75rem for the tabs numbers font size.
- ✅ Let's change the contact name font weight to 400.
- ✅ The create payment request button should have 12px of padding top and bottom.
- ✅ The title and the content in the op ups is not aligned
- ✅ The 3 dots should look like they have the same distance with the edge of the card and with the EUR text

---
Will investigate over:
- 🧐 After creating a payment request, I can't scroll anymore on the list (I don't know if this is a problem with Storybook).

Will wait for designs for:
- 🎨 We should redesign the pager, it's too small to be used on mobile.
- 🎨 I'm making new payment methods icons to adapt better to this size.

PR #89, opened 2023-07-13
[MOOV-2010]: Added totals per currency on PR tabs

agranillonf:
Added total amount per currency on the Payment Request Status filter tabs.

| Mobile      | Tablet       | Desktop   |
| ----------- | ----------- |------------|
|image|image|image|


arifmatin (approved):

pabloi-nf (changes requested):
image

- The amount should show decimals, and thousand separators.
- There is an error when creating a PR from the platform, when the amount is not being updated.
- I first tried this in prod, and this error happen. Donal says to mention that the code shouldn't fail if the totals aren't returned.
image

pabloi-nf (changes requested):
In the "paid" tab is not showing the value in pounds. I understand it's because the value is zero, but it can create some confusion to don't see the value while you can see it in other tabs. I agree that if both values are 0 (like in the partial payments) it's OK to don't show the amounts. So maybe we show both values unless there is no PR's at all.

image

agranillonf:
@pabloi-nf Totals per currency and status will now only show if:
- There are payment requests for that specific status
- There are payment requests for that specific currency, no matter the status (E.g. within the "All" tab).

agranillonf:
| Complete totals | Partial totals |
|-|-|
|image|image|

donalnofrixion (approved):

Publicker:
> Complete totals	Partial totals
> image	image

Based on @pabloi-nf's feedback, I believe that if at least one currency is shown, we'd show both. So using the image as an example, the "Partial totals" scenario, "All", "Partially paid", and "Paid" tabs, should show "£0".

@pabloi-nf Did I get that wrong?

pabloi-nf (approved):

Publicker (approved):

PR #95, opened 2023-07-20
[MOOV-1970]: Fixed PR outstanding amount adding authorized PISPs

agranillonf:
Partially paid PRs will now only show the sum of completed PISP payments in the outstanding balance component.


Publicker (approved):

donalnofrixion (approved):

pabloi-nf (changes requested):
We need to show something under the completion line. Let's use the same format "€X.XX outstanding", but with the total amount.

image

agranillonf:
@pabloi-nf We're showing that legend but only when there has been a successful payment. With these changes, because Authorized PISP payments are no longer considered paid, the legend won't show. Should we show it whenever there's a transaction on the list?

pabloi-nf:
@agranillonf yes, let's show it for all PRs with partial payments enabled, no matter how much was actually paid

sipsorcery (approved):

PR #99, opened 2023-07-26
[MOOV-2029]: Removed dynamic initialization of defaults checkbox

agranillonf:
Disabled dynamic initialization of User Defaults modal checkbox, so that it doesn't preserve its checked state between payment request creations.


donalnofrixion (approved):

arifmatin (approved):

pabloi-nf (changes requested):
It doesn't save the defaults when I try to make another PR.
In the video can be seen that I leave only PIS as the payment method, check the checkbox and Apply. When I try to create another PR the settings are not saved.

https://github.com/nofrixion/nofrixion.webcomponents/assets/127946572/0991438c-fe9f-46d4-9df6-f9f53564bbc2

agranillonf:
@pabloi-nf Settings were actually being saved but not reloaded, which is fixed now.

arifmatin (approved):

pabloi-nf (changes requested):
Now it works well!
One last comment: I notice it's not helping me to know if I'm making changes or not to my previous defaults. A possibility is to only show the "use as my default" checkbox if the current configuration is different than the saved one. Is that possible?

agranillonf:
That's a great idea @pabloi-nf ! We'll try and update this PR when ready.

Publicker:
Hey @agranillonf, I've noticed some strange behaviours while using it.

1. While changing the Payment Methods, I could see that I changed two settings (don't capture and defining a priority bank), and in both changes the "Use as my default" didn't appear.
2. Whenever I typed the Payment notifications email, the "Use as my default" will appear and disappear constantly.

Showcase video:

https://github.com/nofrixion/nofrixion.webcomponents/assets/52673485/3b74eb5e-d7ec-47dd-adc8-8786c3cf245c

pabloi-nf (commented):
I noticed the same as @Publicker. It seems it's appearing and dissapearing randomly sometimes. 
Maybe it has to do with the first use and not having defaults saved? If that's the case, the first the user opens this boxes the checkbox should be enabled, because there is no defaults.
The checkbox should be hidden only if the saved settings match the current settings.

agranillonf:
@Publicker @pabloi-nf Could you try again with the latest version?

Publicker:
@agranillonf Things are looking much better now, but I just found one thing that doesn't behave as expected.

When you enable the definition of a priority bank, the 'Use as my default' option doesn't come up, even though the new selected value doesn't match the defaults. The modals only recognize the difference with the defaults and show the 'Use as my default' checkbox when changing the bank. Also, after changing that bank, the checkbox doesn't disappear until you tap the cross, instead of applying it.

Showcase video:

https://github.com/nofrixion/nofrixion.webcomponents/assets/52673485/081f8c79-9118-4ea6-bc67-1c8e5021a64c

agranillonf:
@Publicker @pabloi-nf Issue has been fixed. Could you try again, please?

Publicker (approved):
Yep, working smoothly now! Thanks @agranillonf

pabloi-nf (changes requested):
I still get an error if I leave the notfications field empty and check it as default

agranillonf:
@pabloi-nf My bad, the other PR wasn't merged yet. However, it is now. Can you try again, please?

pabloi-nf (approved):

PR #106, opened 2023-08-01
Initial refund functionality

sauravmaiti22:
- Added refund modal which supports partial refunds.
- If the card is authorised but not captured, partial refund is not possible.
- Added a paymentAttemptsHelper.ts which contains calculations in a common place instead of scattered everywhere.
- Have tested a few scenarios and needs further testing.
- WIP: Need API changes to support scenarios like partial capture and refund. For now, only try happy paths.


donalnofrixion (approved):

src/components/ui/CardRefundModal/CardRefundModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  currency: Currency.EUR | Currency.GBP;
+  onRefund: () => Promise<void>;
+  onDismiss: () => void;
+  setAmountToRefund: (amount: string) => void;
+  maxRefundableAmount: number;
+  lastFourDigitsOnCard?: string;
+  processor?: string;
+  transactionDate: Date;
+  contactName?: string;
+  IsPartialRefundPossible: boolean;
```
Publicker, line 20: Starting a variable with uppercase should be reserved only for components

agranillonf (approved):

sipsorcery (approved):

pabloi-nfxn (changes requested):
The icon should be smaller. It's in 16px, and it should be in 12px.
image
image

pabloi-nfxn (approved):

PR #110, opened 2023-08-03
[MOOV-1989]: Added warning when only enabling wallet payments

agranillonf:
When creating a payment request, if the user only enables Apple Pay / Google Pay, a warning telling them to make sure their users are able to use those platforms is shown under the payment methods.


donalnofrixion (approved):

Publicker (approved):

pabloi-nf (changes requested):
Works perfect.
Let's update the text to:
"Do your customers have access to Apple Pay or Google Pay? If you are unsure, you may want to consider adding a second payment method as a backup."

pabloi-nf (approved):

